PR #24, opened 2019-01-25
feat(LOC-221): add Avatar component

flycrum:

**Summary:** Added an `Avatar` component with image and initials visuals, team and user types, background colors, and sizes. To support this feature, a `ClippedContent` component has been created and the `ImageCircle` was modified to use `ClippedContent`.

**Screenshots:** 
image
image
image

### DIFF
--- a/src/components/ImageCircle/ImageCircle.sass
+++ b/src/components/ImageCircle/ImageCircle.sass
@@ -1,6 +1,6 @@
 @import '../../styles/_partials/index'
 
-body .ImageCircleContainer
+.ImageCircleContainer
 	width: 40px
 	min-width: 40px
 	height: 40px
--- a/src/components/ImageCircle/ImageCircle.sass
+++ b/src/components/ImageCircle/ImageCircle.sass
@@ -1,6 +1,6 @@
 @import '../../styles/_partials/index'
 
-.ImageCircleContainer
+body .ImageCircleContainer
 	width: 40px
 	min-width: 40px
 	height: 40px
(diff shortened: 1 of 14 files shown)


src/components/ImageCircle/ImageCircle.sass
```diff
@@ -1,6 +1,6 @@
 @import '../../styles/_partials/index'
 
-.ImageCircleContainer
+body .ImageCircleContainer
```
flycrum: This is an unfortunate workaround having to deal with `styles` added by a developer. There's certainly a better solution, but I didn't want it to slow down progress.

src/components/ImageCircle/ImageCircle.tsx
```diff
@@ -22,17 +26,20 @@ export default class ImageCircle extends React.Component<IProps> {
 	};
 
 	render () {
-		const ContainerTag: any = this.props.tag;
-
 		return (
-			<ContainerTag
+			<ClippedContent
```
flycrum: Now using `ClippedContent` internally so its consistent sizing and shape with Avatar initials.

src/components/VerticalNav/VerticalNav.tsx
```diff
@@ -145,6 +146,7 @@ interface IWorkspaceSwitcherProps extends IReactComponentProps {
 	onClickAddTeam: Handler;
 	onClickLogout: Handler;
 	onClickWorkspace: Handler;
+	onClickWorkspaceNav: Handler;
```
flycrum: This eventual is specifically for when the `VerticalNav` icon is clicked.

src/components/ImageCircle/ImageCircle.sass
claygriffiths: Yeah, I'd like to see this eliminated if possible.

flycrum: Good call. Done. Added some props to remove the need for this.

PR #25, opened 2019-01-28
build(LOC-336): output .d.ts for consumers of local-components, remove react-hot-loader

claygriffiths:
**Summary:** Output `.d.ts` for consumers of `local-components` such as Local itself in addition to Add-ons and Environments. Doing so allows IDEs to pull in the component props and types.

Unfortunately, the way we integrated `react-hot-loader` would not play nicely with outputting TypeScript declarations so it has been removed for the time being.

**Other Notable Changes:**

* Fixed issue with how `react-truncate-markup` was loaded
* Fixed issue where SASS changes would trigger a refresh in Styleguidist rather than use HMR

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
 		"react": "^16.4.2",
 		"react-docgen-typescript": "^1.12.2",
 		"react-dom": "^16.4.2",
-		"react-styleguidist": "8.0.5",
+		"react-styleguidist": "^8.0.6",
 		"resolve-url-loader": "^3.0.0",
 		"sass-loader": "^7.1.0",
 		"shx": "^0.3.2",
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -1,6 +1,7 @@
 const webpack = require('webpack');
 const path = require('path');
 const merge = require('webpack-merge');
+const package = require('./package.json');
 
 module.exports = {
     components: 'src/**/[A-Z]*.tsx',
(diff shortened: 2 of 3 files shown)


src/components/TertiaryNav/TertiaryNav.tsx
```diff
@@ -6,13 +6,18 @@ import * as styles from './TertiaryNav.sass';
...
-interface IProps extends IReactComponentProps {
+interface ITertiaryNavProps extends IReactComponentProps {
 
-	children?: Array<React.ReactElement<TertiaryNavItemBase>> | React.ReactElement<TertiaryNavItemBase>;
+	children?: Array<React.ReactElement<TertiaryNavItem>> | React.ReactElement<TertiaryNavItem>;
 
 }
 
-class TertiaryNavBase extends React.Component<IProps & IReactComponentProps & RouteComponentProps<{}>> {
+// There is a known issue in TypeScript, which doesn't allow decorators to change the signature of the classes
```
flycrum: This is interesting. Do you consider this to simply be a minor annoyance?

src/components/Truncate/Truncate.tsx
```diff
@@ -1,9 +1,7 @@
 import * as React from 'react';
 import IReactComponentProps from '../../common/structures/IReactComponentProps';
 import Omit from '../../common/structures/Omit';
-
-/* TruncateMarkup's types and exports are all jacked up. This makes it play nicely in Local. */
-const TruncateMarkup = require('react-truncate-markup').default;
+import * as TruncateMarkup from 'react-truncate-markup/lib';
```
flycrum: So weird how they set this library up. Glad this resolves the issue, however. Great job!

tsconfig.json
```diff
@@ -4,7 +4,9 @@
 		"module": "commonjs",
 		"lib": ["es2017", "dom", "dom.iterable"],
 		"sourceMap": true,
+		"declaration": true,
 		"strict": true,
+		"experimentalDecorators": true,
```
flycrum: 🎉

PR #26, opened 2019-02-05
improvement(Workspaces): add placeholder img and initials color to Avatar component

flycrum:
**Audience:** Users | Engineers | Third-party Developers

**Summary:** Refactor Avatar component to include placeholder img, add support for a background color for initials, fix border around both selected and unselected, and rework the layout so any initials always appear below the placeholder and primary img in the event that it takes awhile to load, is a transparent png, or silently fails.

**Technical:** Now using `position: absolute` to show multiple elements on top of one another at the same time. This PR addresses LOC-198 and LOC-353.

**Screenshots:**
image
image


src/components/ImageCircle/ImageCircle.tsx
```diff
@@ -48,9 +48,6 @@ export default class ImageCircle extends React.Component<IProps> {
 					className={classnames(
 						styles.ImageCircle,
 						this.props.className,
-						{
```
flycrum: This wasn't being used.

PR #27, opened 2019-02-06
improvement(LOC-320): add text support to Spinner, use onRequestClose prop in FlyModal

claygriffiths:
**Summary:** Add text/children support for `<Spinner />`, use `onRequestClose` prop on `<FlyModal />` if provided. Some SCSS was also removed that should've been in Local itself.

----

### Screenshot

screen shot 2019-02-06 at 9 54 15 am

### DIFF
--- a/src/components/FlyModal/FlyModal.tsx
+++ b/src/components/FlyModal/FlyModal.tsx
@@ -48,7 +48,6 @@ export default class FlyModal extends React.Component<IProps> {
 		closeTimeoutMS: 0,
 		hasIcon: false,
 		isOpen: true,
-		onRequestClose: FlyModal.onRequestClose,
 		overlayClassName: styles.FlyModalOverlay,
 		parentSelector: () => document.body,
 		portalClassName: 'ReactModalPortal',
@@ -60,10 +59,19 @@ export default class FlyModal extends React.Component<IProps> {
 		ReactDOM.unmountComponentAtNode(document.getElementById('popup-container'));
 	}
 
+	onRequestClose = () => {
+		if (typeof this.props.onRequestClose === 'function') {
+			this.props.onRequestClose();
+		}
+
+		FlyModal.onRequestClose();
+	};
+
 	render () {
 		return (
 			<ReactModal
 				{...this.props}
+				onRequestClose={this.onRequestClose}
 				className={classnames(
 					styles.FlyModal,
 					'FlyModal', // in here for tests
@@ -73,7 +81,7 @@ export default class FlyModal extends React.Component<IProps> {
 					className={classnames({
 						[styles.FlyModal__HasIcon]: this.props.hasIcon,
 					})}
-					onClick={FlyModal.onRequestClose}
+					onClick={this.onRequestClose}
 				/>
 				{this.props.children}
 			</ReactModal>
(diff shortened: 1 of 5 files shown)


src/components/FlyModal/FlyModal.tsx
```diff
@@ -33,7 +33,7 @@ interface IProps extends IReactComponentProps {
 	closeTimeoutMS?: number;
 	hasIcon?: boolean;
 	isOpen?: boolean;
-	onRequestClose?: () => void;
+	onRequestClose?: () => boolean | void;
```
flycrum: Just curious, why is this boolean or void?

src/components/Spinner/README.md
```diff
@@ -1,4 +1,11 @@
-Basic spinner.
+Spinner without text.
 ```js
 <Spinner />
 ```
+
+Spinner with text.
+```js
+<Spinner>
+	With some text...
```
flycrum: I love seeing this proved out in Styleguidist! 🙌

src/components/Spinner/Spinner.tsx
```diff
@@ -5,10 +5,8 @@ import SpinnerSVG from '../../svg/spinner';
 import * as styles from './Spinner.sass';
 
 interface IProps extends IReactComponentProps {
-
-	className?: string;
-	color?: 'Gray25' | 'GrayDark50';
-
+	className?: string
```
flycrum: Yeah, I'm cool with ditching these extra lines. If we still want to keep them for classes, I think that's fine, but I'd love to ditch them for everything else.

src/styles/containers/SiteInfo.scss
```diff
@@ -333,28 +333,4 @@
 			fill: $gray75;
 		}
 	}
-
```
flycrum: Bye bye Local-specific styles! 👋

src/components/FlyModal/FlyModal.tsx
```diff
@@ -73,7 +81,7 @@ export default class FlyModal extends React.Component<IProps> {
 					className={classnames({
 						[styles.FlyModal__HasIcon]: this.props.hasIcon,
 					})}
-					onClick={FlyModal.onRequestClose}
+					onClick={this.onRequestClose.bind(this)}
```
flycrum: This is the only thing I'd really like to see changed. I'd prefer avoid ever using bind within the `render` function as it some undesirable re-rendering implications.

src/components/FlyModal/FlyModal.tsx
claygriffiths: Good catch. I initially made it so returning false on `onRequestClose` would stop the modal from closing but I'm going to ditch that for now.

flycrum: That makes sense. I wondered cause I've done the same thing when building modal components in the past. Just didn't see it being used and didn't know if there were plans to use it that way later on.

src/components/Spinner/Spinner.tsx
claygriffiths: Yeah, we really need to get some TSLint rules in place like we did with ESLint. It's a shame there isn't an easy way to port the settings.

PR #33, opened 2019-03-14
improvement(LOC-505): add vertical layout to RadioBlock

flycrum:
**Audience:** Users | Developers

**Summary:** Added direction prop and vertical layout/styles to the Radio Block component. An example has been added to Styleguidist as well.

**Screenshots:** 

image


adamtrabold:
More detailed context as per Chris' request –

The total expansion should be 4px on each side, from the center, including the border. Sketch attached.
Image from iOS (7)

PR #34, opened 2019-04-03
feat(LOC-556): create Container component with advanced margin expressions and auto disable feature

flycrum:

**Summary:** Most components within the `local-components` package are wrapped in a container to allow for more flexibility (which allows for more style and layout flexibility while being able to more gracefully handle unplanned edge-cases). `Container` provides a predictable wrapper component, with a consistent API, and convenient way to access the Container's "shape" through a single `component` prop within the component utilizing it (e.g. Header).

The `Container` component also adds some extremely helpful and convenient props/features. One of those features is to disable/exclude the container DOM element if not being explicitly used. This is useful if you want the flexibility of using the container if needed without creating an additional element when not used.

The most dynamic feature of the `Container` component is its margin expression system. This will allow developers to layout a component (that's wrapped in a Container) to set its own margin without having to create a CSS rule and apply it to `className`. This is particularly useful and practical for vertical layouts where there are a number of elements, one after another, that need various margins set to create whitespace between them. While the traditional approach would to either bake those margins into the original component to follow existing patterns (inflexible and assumption-based) or to create CSS for a group or individual elements, neither really provides 1) consistency, 2) convenience, and 3) flexibility.

The margin system built within the `Container` component attempts to solve all three:
1. Consistency: the margin system utilizes existing margin sizes used throughout Local (e.g. xs, s, m, l, xl) and makes those available to developers. And if applied to every component within the `local-components` ecosystem, this consistency and predictability would greatly simplify UI implementation.
2. Convenience: the margin system allows develops to set margins directly on the component they're implements. No longer do developers need to write CSS rules just to create whitespace between elements (which tends to needlessly pollute CSS stylesheets with static and otherwise meaningly px values)
3. Flexibility: the margin system uses expressions that allow developers to handle nearly any situation imaginable. Need to utilize the `m` margin sizing but the component has an extra couple of pixels worth of padding around it thus throwing off the whitespace no longer making it appear like a `m` margin size worth of spacing? Simply use an expression like `m-2` to account for this otherwise painful edge-case.

**Examples:**

Usage examples of the margin system:

_simple margin sizing (result: `margin: 20px;`)_
```
<Header container={{margin: 'm'}}>
```
_margin shorthand sizing (result: `margin: 20px 5px 10px;`)_
```
<Header container={{margin: 'm xs s'}}>
```
_mixing margin shorthand sizing and basic arithmetic (result: `margin: 18px 25px;`)_
```
<Header container={{margin: 'm-2 m+s'}}>
```
_mixing margin-bottom and explicit margin-left (result: `margin-bottom: 20px; margin-left: 13px;`)_
```
<Header container={{marginBottom: 'm', marginLeft: '13px'}}>
```

**Out of scope:** All components outside of `Header`. This is a proof of concept that needs to be accessed and determined whether it should be applied across the `local-components` family of components.

### DIFF
--- a/src/components/Container/Container.scss
+++ b/src/components/Container/Container.scss
@@ -1,5 +1,0 @@
-@import '../../styles/_partials/index';
-
-.Container {
-
-}
--- a/src/components/Container/Container.scss
+++ b/src/components/Container/Container.scss
@@ -1,0 +1,5 @@
+@import '../../styles/_partials/index';
+
+.Container {
+
+}
--- a/src/components/Container/ContainerMarginHelper.test.ts
+++ b/src/components/Container/ContainerMarginHelper.test.ts
@@ -1,0 +1,81 @@
+import { ContainerMarginHelper } from './ContainerMarginHelper';
+
+it ('test expressions lookups', () => {
+	expect(ContainerMarginHelper.parseExpression('xs')).toBe(ContainerMarginHelper.sizeLookups.get('xs'));
+	expect(ContainerMarginHelper.parseExpression('xs')).not.toBe(ContainerMarginHelper.sizeLookups.get('s'));
+	expect(ContainerMarginHelper.parseExpression('s')).toBe(ContainerMarginHelper.sizeLookups.get('s'));
+	expect(ContainerMarginHelper.parseExpression('m')).toBe(ContainerMarginHelper.sizeLookups.get('m'));
+	expect(ContainerMarginHelper.parseExpression('l')).toBe(ContainerMarginHelper.sizeLookups.get('l'));
+	expect(ContainerMarginHelper.parseExpression('xl')).toBe(ContainerMarginHelper.sizeLookups.get('xl'));
+	expect(ContainerMarginHelper.parseExpression('0')).toBe(ContainerMarginHelper.sizeLookups.get('0'));
+	expect(ContainerMarginHelper.parseExpression('none')).toBe(ContainerMarginHelper.sizeLookups.get('none'));
+});
+
+it ('test expressions negative lookups', () => {
+	expect(ContainerMarginHelper.parseExpression('-m')).not.toBe(ContainerMarginHelper.sizeLookups.get('m'));
+	expect(ContainerMarginHelper.parseExpression('-m')).toBe(-ContainerMarginHelper.sizeLookups.get('m')!);
+});
+
+it ('test expressions multiple lookups', () => {
+	expect(ContainerMarginHelper.parseExpression('m+m')).toBe(ContainerMarginHelper.sizeLookups.get('m')! + ContainerMarginHelper.sizeLookups.get('m')!);
+	expect(ContainerMarginHelper.parseExpression('m-m')).toBe(ContainerMarginHelper.sizeLookups.get('m')! - ContainerMarginHelper.sizeLookups.get('m')!);
+	expect(ContainerMarginHelper.parseExpression('-m-m')).toBe(-ContainerMarginHelper.sizeLookups.get('m')! - ContainerMarginHelper.sizeLookups.get('m')!);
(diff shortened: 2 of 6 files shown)


src/components/Container/Container.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../styles/_partials/index';
+
+.Container {
```
claygriffiths: Is it worth keeping this around?

src/components/Container/ContainerMarginHelper.test.ts
```diff
@@ -0,0 +1,81 @@
...
+});
+
+it ('test invalid expressions mixed with valid', () => {
+	expect(ContainerMarginHelper.parseExpression('zzz+1')).toBe(1);
+	expect(ContainerMarginHelper.parseExpression('-zzz-1')).toBe(-1);
+	expect(ContainerMarginHelper.parseExpression('-zzz-m+1')).toBe(-ContainerMarginHelper.sizeLookups.get('m')! + 1);
+	expect(ContainerMarginHelper.parseExpression('-zzz+m+1')).toBe(ContainerMarginHelper.sizeLookups.get('m')! + 1);
+	expect(ContainerMarginHelper.parseExpression('1+zzz-zzz+1')).toBe(2);
+	expect(ContainerMarginHelper.parseExpression('-1zzz+zzz-1')).toBe(-2);
+});
```
claygriffiths: Great job on these!

src/components/Container/ContainerMarginHelper.ts
```diff
@@ -0,0 +1,120 @@
+import { IContainerProps } from './Container';
+
+export type ContainerMarginType = 'xs' | 's' | 'm' | 'l' | 'xl' | 0 | 'none' | string;
+type ReturnValue =  {margin?: string | number} | undefined;
+
+
+export class ContainerMarginHelper {
+
+	public static sizeLookups = new Map<string | number, number>(Object.entries({
```
claygriffiths: This is a clever way to get around the initialization format with `Map` 😃

src/components/Container/ContainerMarginHelper.ts
```diff
@@ -0,0 +1,120 @@
+import { IContainerProps } from './Container';
+
+export type ContainerMarginType = 'xs' | 's' | 'm' | 'l' | 'xl' | 0 | 'none' | string;
+type ReturnValue =  {margin?: string | number} | undefined;
```
claygriffiths: Is this type compatible with props like `marginTop`? Should it be something like `{[margin: 'margin'|'marginTop'|...]: string | number}`?

Just caught my eye, I may be way off here.

src/components/Container/ContainerMarginHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+		return this.sizeLookups.get(value);
+	}
+
+	public static wrapReturnValue (value: string | number | undefined, propName: string): ReturnValue {
+		if (value === undefined) {
+			return undefined;
+		}
+
+		return {
+			[propName]: value,
```
claygriffiths: See comment above about the `ReturnValue` type.

src/components/Container/Container.scss
flycrum: No. I'll delete it. If there's ever a time in the future when we want it, I'll add it then.

src/components/Container/ContainerMarginHelper.ts
flycrum: Good catch! I totally missed updating that.

src/components/Container/ContainerMarginHelper.ts
flycrum: Yeah, I went ahead and added more typings to help prevent missing this going forward 👍

PR #35, opened 2019-04-03
improvement (LOC-557): create magic sync viewer style alternatives for existing components

flycrum:
**Audience:** Engineers

**Summary:** Adds styles necessary for the Magic Sync Viewer.

**Technical:** Changes include:
- updated disabled styles
- new readonly styles
- fixed background styles
- disable rendering of Checkbox label element if no label is provided
- add static positioning option for Close component
- set `props.style` and `props.className` throughout various components
- add alternative light and dark mode styles
- change `box-shadow` borders to `border` for several components


src/styles/_partials/_selectors.scss
```diff
@@ -61,3 +61,35 @@
 		@content;
 	}
 }
+
+@mixin selectors_parent {
```
claygriffiths: You and your SASS 😂

PR #36, opened 2019-04-03
feat(LOC-506): add VirtualList and VirtualTable components

flycrum:

**Summary:** There's a need to be able to render lists and tables with large datasets in a manner that scales and performs well whether there be 100 rows or 100,000. Interactive Styleguidist examples have been provided for both VirtualList and VirtualTable.

VirtualList is a largely unstyled series of visible and overscan rows that are virtually swapped in and out while being positioned within a statically set scroll area that represents the size of all items (as if they were being rendered in their entirety).

VirtualTable builds on VirtualList while adding table concepts and Local-specific styles.

**Technical:** 
- both VirtualList and VirtualTable require rows heights to be consistent and predictable
- unlike most virtual list components available, VirtualList and VirtualTable don't use absolute positioning but rather takes advantage of the static layout flow to allow for greater flexibility, resizing animations (potentially), and scrolling capabilities.
- VirtualTable can accept an array of objects or a multi-dimensional array for its `data` prop
- react's render prop pattern is used throughout both components to allow for custom styles, cell contents, and handling of inputs
- memoize is being used to calculate virtualized items accurately and efficiently
- resize-observer is used to determine initial and changing component height

**Screenshots:**
image
image


src/components/VirtualList/README.md
```diff
@@ -0,0 +1,188 @@
+Render large lists by using virtualization to draw only the visible and overscan items.
+
+Configurations:
+
+```js
+class Example extends React.Component {
```
claygriffiths: I wonder if it's worth extracting out these heavy examples into dedicated example components; especially considering how complex the render method is.

PR #38, opened 2019-04-05
feat(LOC-503): add MagicSync Viewer UI/UX

flycrum:
**Audience:** Users | Engineers

**Summary:** The design review for the Magic Sync Viewer which will give Local Team users the power to selectively choose which files they pull or push to Flywheel staging or production environments.

**Screenshots:**
image
image
image
image

### DIFF
--- a/src/components/MagicSyncViewer/README.md
+++ b/src/components/MagicSyncViewer/README.md
@@ -1,4 +1,4 @@
-Interactive Magic Sync Viewer Example:
+Interactive Magic Sync Viewer:
 
 ```js
 const { MagicSyncViewerExample } = require('./examples/MagicSyncViewerExample');


adamtrabold:
This is looking really really close! Only 2 small ui things related **only** to Dark Mode --

- Folder icon outline should be #c7c4c4
- Folder icon interior should be #5d5e5e (transparent when opened)
- Checked/unchecked text should match existing checkbox component. I'm not sure if there's an error in the component, or just this screen...when the box is checked, text should be #fff @ 500 weight...when unchecked, #e7e7e7 @ 300 weight.

Otherwise, LGTM!

adamtrabold:
One other thing - the select should match our inline select, not our standard selects. I don't think this pattern currently is in use in the Local app, but it is in Teams management on FW. Image attached!

Screen Shot 2019-04-05 at 10 23 07 AM

PR #53, opened 2019-06-03
fix: remove filter to fix slow WorkspaceSwitcher animation

flycrum:
**Audience:** Users | Engineers

**Summary:** The WorkspaceSwitcher animation performance was poor on first reveal and sometimes subsequent calls to show and/or hide. This fix removes the filter effect as it was identified as the culprit.

**Technical:** `filter: drop-shadow...` was not playing nice with `transition` so it was replaced with `box-shadow`. Several style instances of transition and box shadow were changed in conjunction with this change. In addition, `will-change` has been added as a performance measure to Popup and Tooltip.

### DIFF
--- a/src/components/FlyTooltip/FlyTooltip.scss
+++ b/src/components/FlyTooltip/FlyTooltip.scss
@@ -45,7 +45,6 @@
 	}
 
 	&::before {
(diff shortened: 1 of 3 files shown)


andypeters:
👀 ✨ 🎱

PR #55, opened 2019-06-27
style(LOC-777): add hover states to workspace switcher and adjust clip-path of avatars

flycrum:
**Audience:** Users | Engineers | Third-Party Developers

**Summary:** The workspace switcher is missing several hover states. This PR adds those styles while also adjusting the clip-path of avatars (via ImageCircle) to address alleged rounding issue that results in an unexpected y-start of the clip path.

**Screenshots:**

Before clip-path fix:
image

After clip-path fix:
image

Inactive item (base vs hover):
image

Active item (base vs hover):
image

Add button (base vs hover):
image

### DIFF
--- a/src/components/ClippedContent/ClippedContent.sass
+++ b/src/components/ClippedContent/ClippedContent.sass
@@ -3,7 +3,7 @@
 .ClippedContent
 	display: flex
 	flex-direction: column
-	clip-path: inset(0 round 4px)
+	clip-path: inset(-1px 0 round 4px)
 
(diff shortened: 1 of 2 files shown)


adamtrabold:
These look fantastic! Added a bug to JIRA about the + icon being squared off instead of rounded. Otherwise, gtg!

flycrum:
> These look fantastic! Added a bug to JIRA about the + icon being squared off instead of rounded. Otherwise, gtg!

Awesome, thanks for adding that!

PR #57, opened 2019-07-26
feat(LOC-1043): add animation modal and list example using react-move

flycrum:
**Audience:** Engineers | Third-Party Developers

**Summary:** For more intricate transitions, a library is needed to simplify the mounting/unmounting of animations that require handling of states beyond simply start and end. This PR introduces examples for lists that animate in and out items as well as a slide in modal animation. These examples are under the menu item `Animation` for the time being.

**Out of Scope:** Components. This PR does not attempt to create a reusable component but rather proves out two scenarios and solutions where a library like `react-move` would be an asset for developers.

### DIFF
--- a/src/components/Animation/Animation.tsx
+++ b/src/components/Animation/Animation.tsx
@@ -1,0 +1,9 @@
+import * as React from 'react';
+
+export default class Animation extends React.Component {
+
+	render () {
+		return this.props.children;
+	}
+
+}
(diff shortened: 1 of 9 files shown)


src/components/Animation/Animation.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export default class Animation extends React.Component {
```
claygriffiths: Is this component still needed?

flycrum: Great question. I couldn't get the example to load up in styleguidist unless there was an actual component. I thought that odd. I am curious about different ways of abstracting some of the inner working of `react-move` for reuse and easy of implementation, but I didn't want to get caught up in that for this.

PR #58, opened 2019-08-20
feat(LOC-1155): create TitleBar component

flycrum:
**Audience:** Third-party developers | Engineers

**Summary:** This adds an out-of-the-box TitleBar component for Add-on developers to easily implement while not having to worry about styles like font-sizes, font-weights, and dark mode text and background colors. The Local app also has several instances that could be refactored to use this component.

**Technical:** `TitleBar` takes a `title` prop as well as children content that will be right-aligned (if provided).

**Out of Scope:** In dark mode, several instances where this component is used have a border above it that's slightly lighter and creates a beveled look. This is not intentional and while a `margin-top: -1px` fix could resolve the issue, it's far from ideal to bake that into the base style of the component and thus has not been implemented.

**Screenshots:**
image

image

note: the font-weight below is not representative of how it will look in the app
image

### DIFF
--- a/src/components/TitleBar/TitleBar.tsx
+++ b/src/components/TitleBar/TitleBar.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import ILocalContainerProps from '../../common/structures/ILocalContainerProps';
+import * as styles from './TitleBar.scss';
+import classnames from 'classnames';
+import { Container } from '../Container';
+
+interface IProps extends ILocalContainerProps {}
+
+export const TitleBar = (props: IProps) => (
+	<Container>
+		<div
+			className={classnames(
+				styles.TitleBar,
+				'TitleBar',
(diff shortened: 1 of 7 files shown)


src/components/TitleBar/TitleBar.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Container } from '../Container';
+
+interface IProps extends ILocalContainerProps {}
+
+export const TitleBar = (props: IProps) => (
+	<Container>
+		<div
+			className={classnames(
+				styles.TitleBar,
+				'TitleBar',
```
flycrum: The idea here is to create an unstyled, globally-scoped selector here (`'TitleBar'`), so it's possible to target the component and apply any edge-case styles. The goal would be to eventually apply this to all our components.

PR #60, opened 2019-08-21
Update Styleguidist Workflow to use macOS and Yarn

claygriffiths:


claygriffiths:
Scrapping GitHub Actions for now in favor of Azure Pipelines.

claygriffiths:
GitHub Actions is still in its infancy so for the sake of saving time, I will be creating a workflow in Azure Pipelines instead.

PR #66, opened 2019-08-29
feat(LOC-1162): add Text, Title, and TextBase components with examples

flycrum:
**Audience:** Engineers | Add-on Developers

**Summary:** Provide first-class text components with a simple API. This PR introduces 2 public components: Title and Text as well as a private TextBase component.

Interactive examples, with all available options, are provided via Styleguidist docs and all other components refactored to use the new buttons.

**Technical:**
- The `Header` component has been removed and all instances replaced with `Title`
- This component has been architected the same way as the newly developed button components
- `TextBase` is for internal purposes. This component's responsibility is to provide an API for other text components to use when creating their own set of styles and variations. It is not the TextBase's responsibility to manage, restrict, or enforce rules itself as it will be the responsibility of public components to do that. One of the huge advantages of this approach is that each of the text components can set their own public API and avoid undesirable scenarios where props exist but can't be used with certain combinations or dev log warnings are thrown to indicate exceptions.
- There are a set of global styles for texts that will continue to be applied. These styles, however, have been moved to their own `text.scss` file. Included in these global styles are base10 font-size, text color, font size, font weight, font family, h1-h6 resets, paragraph, and links.

**Example:** Below is an example of a `Text` component nested within another Text component with a link child to show what's possible with text and how styles remained scoped to each specific instance:

image

image

**Screenshots:**

image

image


src/index.ts
```diff
@@ -27,7 +27,6 @@ export {default as FlyLargeConfirm} from './components/FlyLargeConfirm';
 export {default as FlyModal} from './components/FlyModal';
 export {default as FlySelect} from './components/FlySelect';
 export {default as FlyTooltip} from './components/FlyTooltip';
-export {default as Header} from './components/Header';
```
claygriffiths: Be sure to re-add the new components from this PR to this file.

src/styles/text.scss
```diff
@@ -0,0 +1,46 @@
+@import '_partials/index';
+
+:global {
```
claygriffiths: Good stuff!

src/components/List/List.tsx
```diff
@@ -70,20 +65,19 @@ export default class List extends React.Component<IProps> {
...
+					<Title
 						className={classnames(
 							styles.List_Header,
 							this.props.headerClass,
 							{
 								[styles.List_Header__Divider]: this.props.headerHasDivider,
 							},
 						)}
-						fontSize={this.props.headerFontSize}
-						fontWeight={this.props.headerWeight}
```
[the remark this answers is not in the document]
flycrum: Hmm, I need to do a better job of being aware of these types of implications. I also need to do a better job of documenting breaking changes. I'm going to go back and change that commit name to be `BREAKING CHANGE`.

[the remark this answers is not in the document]
flycrum: But yes, this does impact any Add-on using it and could change the visual appearance of the component. I'd need to do an audit to see if there are any that use it but my hunch is that this is used in Marketplace and no where else.

src/index.ts
flycrum: Yes, this was a bad oversight 😬 Great catch!!

PR #73, opened 2019-09-04
improvement(LOC-983): add styleguidist sections and reorganize components

flycrum:
**Audience:** Add-on Developers | Engineers

**Summary:** Reorganize all components in logical directories and configure Styleguidist to display them in sections. New [Getting Started with Add-ons](https://build.localbyflywheel.com/project/) section that links off to that content also added.

**Technical:** A `Private (no not use)` section has been added for all components within `_private` directories which indicate that they're intended to be used for private/internal component development only. To streamline the refactoring of `README.md` examples, change all `require` usages for components to `import`. This requires Styleguidist to be updated, so it should be updated to `9.1.6` and React and related libraries to `16.9.0`.

**Review Notes:** There are lot of files that changed in this PR. Arguably the key ones are:
- `styleguide.config.js`
- `index.ts`
- some of the examples that now use `import`

**Screenshots:**

image
image


styleguide.config.js
```diff
@@ -18,17 +18,96 @@ module.exports = {
...
+					ignore: '**/_private/**/*.tsx',
+				},
+				{
+					name: 'Inputs',
+					components: 'src/components/inputs/**/[A-Z]*.tsx',
+					ignore: '**/_private/*.tsx',
+				},
+				{
+					name: 'Layout',
+					components: 'src/components/loaders/**/[A-Z]*.tsx',
```
claygriffiths: `Layout` and `Loaders` have the same value for `components`. Is this intentional?

flycrum: Absolutely not. Good catch!

PR #88, opened 2019-11-06
improvement(LOC-1405): add striped row option to FlySelect and add example

flycrum:
## Audience

Users | Local Engineers | Third-Party Developers

## Summary

* Adds a striped/alternate rows option to FlySelect via `striped` prop and adds the corresponding styles.
* Adds a styleguidist example to demonstrate alternate row colors which includes an option groups code example as part of it.

### DIFF
--- a/src/styles/forms.scss
+++ b/src/styles/forms.scss
@@ -129,34 +129,12 @@
 			}
 		}
 
-		.FlySelect_Option {
-			height: 30px;
-			padding: 0 10px;
-
-			.FlySelect__SecondaryText {
-				margin-right: 0;
-			}
-
-			&:hover {
-				color: $white;
-				background: $green-dark50;
-
-				span {
-					color: $white;
-				}
-
-				svg path {
-					fill: $white;
-				}
-			}
-		}
-
 		.FlySelect_OptionGroup {
-			@include theme-background-gray5-else-graydarkalt;
 			height: 37px;
 			line-height: 37px;
 			text-align: left;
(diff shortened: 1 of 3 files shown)


src/styles/forms.scss
```diff
@@ -129,34 +129,12 @@
...
-				span {
-					color: $white;
-				}
-
-				svg path {
-					fill: $white;
-				}
-			}
-		}
-
```
flycrum: This has been moved down below `.FlySelect_OptionGroup` for no reason other than it feels like this would be the order that they'd appear in.

src/styles/forms.scss
```diff
@@ -129,34 +129,12 @@
...
 		.FlySelect_OptionGroup {
-			@include theme-background-gray5-else-graydarkalt;
 			height: 37px;
 			line-height: 37px;
 			text-align: left;
 			@include theme-color-gray-else-gray25;
+			@include theme-background-gray15-else-graydark50;
```
flycrum: Made darker to provide better contrast with striped/alternate rows option and corresponding styles.

src/styles/forms.scss
```diff
@@ -188,6 +166,32 @@
...
+					color: $white;
+				}
+
+				svg path {
+					fill: $white;
+				}
+			}
+
+			&.FlySelect_Option__Striped:nth-child(even):not(:hover) {
+				@include theme-background-gray2-else-gray2row;
```
flycrum: This style matches that of TableList by using the same mixin.

PR #103, opened 2020-05-13
feat(LOC-88): Add tabbing to Add Site

alexasparks:
## Audience

Local Users

## Summary

Users can now navigate through the 'Add Site' forms using only their keyboards.

## Technical

This PR adds onKeyDown event handlers & applies the default Google Chrome focus outlines to the following components:
- Close (styles only, 'Escape' key functionality in `flywheel-local` pr)
- BrowseInput
- FlySelect
- InputPasswordToggle
- RadioBlock
- AdvancedToggle

The story is completed with this [flywheel-local PR](https://github.com/getflywheel/flywheel-local/pull/571)

## Screenshots / Text Samples

BrowseInput
AdvancedToggle
Go Back Button
InputPasswordToggle
Radio Block
Close
Continue
FlySelect

## Reference

- [LOC-88](https://getflywheel.atlassian.net/browse/LOC-88)

### DIFF
--- a/src/components/buttons/Close/Close.tsx
+++ b/src/components/buttons/Close/Close.tsx
@@ -25,7 +25,7 @@ export default class Close extends React.Component<IProps> {
 					{
 						[styles.Close__PositionAbsolute]: this.props.position === 'absolute',
 					},
-					this.props.className
+					this.props.className,
 				)}
 				onClick={this.props.onClick}
 				style={this.props.style}
--- a/src/components/buttons/Close/Close.tsx
+++ b/src/components/buttons/Close/Close.tsx
@@ -6,7 +6,7 @@ import * as styles from './Close.scss';
 import { FunctionGeneric } from '../../../common/structures/Generics';
 
 interface IProps extends IReactComponentProps {
-	position?: 'absolute' | 'static',
+	position?: 'absolute' | 'static';
 	onClick: FunctionGeneric;
 }
 
@@ -16,9 +16,16 @@ export default class Close extends React.Component<IProps> {
 		position: 'absolute',
 	};
 
+	onKeyDown = (event: any) => {
+		if (event.key === 'Enter'){
+			event.target.click();
+		}
+	}
+
 	render () {
 		return (
 			<span
+				tabIndex={0}
 				className={classnames(
 					styles.Close,
 					'Close', // this also needs to be globally accessible so other component styles can reference it
@@ -28,6 +35,7 @@ export default class Close extends React.Component<IProps> {
 					this.props.className,
 				)}
 				onClick={this.props.onClick}
+				onKeyDown={this.onKeyDown}
 				style={this.props.style}
 			>
 				<CloseBigSVG />
--- a/src/components/buttons/_private/ButtonBase/ButtonBase.scss
+++ b/src/components/buttons/_private/ButtonBase/ButtonBase.scss
@@ -79,6 +79,10 @@ $disabled-text-dark: $gray-dark50;
 		}
 	}
 
+	&:focus {
+		@include defaultOutline;
+	}
+
 	// if default 'button' tag (otherwise don't impose display style)
 	@at-root button#{&} {
 		display: inline-block;
--- a/src/components/buttons/_private/ButtonBase/ButtonBase.scss
+++ b/src/components/buttons/_private/ButtonBase/ButtonBase.scss
@@ -13,8 +13,6 @@ $disabled-text-dark: $gray-dark50;
 	// base styles
 	position: relative;
 	background: transparent;
-	text-transform: uppercase;
-	font-weight: 700;
 	transition: transform 0.1s ease 0s;
 	text-decoration: none;
 	text-align: center;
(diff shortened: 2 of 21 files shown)


src/components/buttons/_private/ButtonBase/ButtonBase.scss
```diff
@@ -79,6 +79,10 @@ $disabled-text-dark: $gray-dark50;
 		}
 	}
 
+	&:focus {
+		@include defaultOutline;
+	}
+
```
flycrum, line 85: I think it might be worth calling out both in this PR (and the product review testers later) that the focus styles are specifically global and affect every instance of these components even outside the scope of this work (the add site form).

src/components/inputs/FlySelect/FlySelect.tsx
```diff
@@ -175,6 +189,63 @@ export default class FlySelect extends React.Component<IProps, IState> {
...
+				open = true;
+				if (focusedIndex < Object.keys(this.state.optionsFormatted).length - 1){
+					focusedIndex++;
+				}
+				break;
+			case 'Tab':
+				open = false;
+				focusedIndex = 0;
+				break;
+		}
```
flycrum, line 215: Fantastic, fantastic work on this component! So excited to see all of this available throughout the app 🙌 

I think the only thing I'd like to eventually figure out is the `space` key. For the default browser select element, the spacebar can also be used to open and select options.

src/components/inputs/FlySelect/FlySelect.tsx
```diff
@@ -175,6 +189,63 @@ export default class FlySelect extends React.Component<IProps, IState> {
...
+			if (!domNode || !domNode.contains(event.target)) {
+				this.setState({
+					focus: false,
+					open: false,
+				});
+			}
+		}
+		catch (error) {}
+	}
+
```
flycrum, line 248: So awesome including this!!!

src/components/buttons/Close/Close.tsx
```diff
@@ -16,18 +16,26 @@ export default class Close extends React.Component<IProps> {
...
+	onKeyDown = (event: any) => {
+		if (event.key === 'Enter'){
+			event.target.click();
+		}
+	}
+
 	render () {
 		return (
 			<span
+				tabIndex={0}
```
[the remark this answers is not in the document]
flycrum: On second thought, I wonder if we want to add an optional prop for this and, if undefined, don't set `tabIndex` at all. Put it on the implementation. Plus, there are instances where using no tabIndex might be desirable here.

src/components/layout/AdvancedToggle/AdvancedToggle.tsx
```diff
@@ -38,6 +39,12 @@ export default class AdvancedToggle extends React.Component<IProps, IState> {
 		});
 	}
 
+	onKeyDown (event: any) {
+		if (event.key === 'Enter' || event.key === ' '){
+			this.toggleAdvanced();
+		}
+	}
+
```
flycrum, line 47: Really nice work on this component, as well! I was worried about this one but this is super slick.

src/components/inputs/FlySelect/FlySelect.tsx
alexasparks: It's pretty easy to add `space` key functionality. I'll include that with my updates!

src/components/inputs/FlySelect/FlySelect.tsx
claygriffiths: Curious, what issues did you run into with `onBlur` that necessitated `onClickOutside`?

src/components/buttons/Close/Close.tsx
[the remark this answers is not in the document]
alexasparks: So, the MDN docs strongly advise against setting tab indices greater than 1 in favor of organizing markup in the order you want the tabs to go. So, instead of messing with tabIndex props for all, I went with the following solutions. Let me know if you disagree with the approach: 

1. For the RadioBlock, I set the tabIndex to -1 when the item is selected, that way the focus goes immediately to the next selectable option.

2. For the buttons, I changed the order of the JSX for the 'Go Back' buttons and 'Continue' buttons, which gave the desired tab order of Form inputs first, then 'Continue', then 'Go Back'.

src/components/inputs/FlySelect/FlySelect.tsx
alexasparks: So, I originally added that because the onBlur handler was called on the container div. So when I tried to add the open/close functionality, since the focus shifted to whatever option was navigated to, the blur handler was called so the options would immediately disappear.

src/components/buttons/Close/Close.tsx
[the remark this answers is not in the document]
flycrum: Awesome! I was wondering if we could reorder those elements and didn't know how feasible that'd be. Looks like you pulled it off. Nice work!

src/components/inputs/FlySelect/FlySelect.tsx
flycrum: Really appreciate it! Just gave it a test drive and it felt great with the space key in there!

src/components/inputs/FlySelect/FlySelect.tsx
claygriffiths: Aha, makes sense!

Maybe something like https://stackoverflow.com/a/60094794 would simplify this?

flycrum: Oh wild, I hadn't seen a solution like that before

alexasparks: That is a good option, but I don't think it's necessary. I removed the onClickOutside code, and moved the blur handler from the select container div to the options container div. That mitigated the issue with just 1 line of code added to the blur handler toggling the open/close state.

PR #104, opened 2020-05-15
fix(PRO-202): allow textTransform to be optional

veryspry:
Creating this as a mandatory `prop` breaks `yarn build:webpack` and additionally, this shouldn't be required. Let's change that.


alexasparks:
Can you remove the console.log statement from this file?

PR #110, opened 2020-05-22
Z-Index tweak to ToolTip component

coulterpeterson:


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@getflywheel/local-components",
-	"version": "13.2.2",
+	"version": "13.2.3",
```
veryspry, line 3: No need to bump the `version` number here -- This gets bumped once we cut an official release. @flycrum could speak more to this process.

flycrum: This is a great point, @veryspry! We'll take care of the version number after the PR is merged/closed, so yes, it'd be best if this was removed. Thanks.

src/common/structures/IState.ts
```diff
@@ -0,0 +1,5 @@
+export default interface IState {
+	addingItem: boolean;
+	initialData: any;
+	unsavedData: any;
+}
```
flycrum, line 5: I'm not sure what this is for but it doesn't appear to be "common". I'd like to see this removed if it's not used for anything and moved to the specific component that uses it if that's the case.

src/components/overlays/Tooltip/Tooltip.scss
```diff
@@ -29,6 +29,7 @@ $transitionDuration: 100ms;
 
 .Tooltip_Popper {
 	visibility: hidden;
+	z-index: 500;
```
flycrum, line 32: Yeah, this makes a lot of sense to me. Thanks for the change!

src/index.ts
```diff
@@ -1,4 +1,4 @@
-import addStyles from './add-styles';
+import addStyles from "./add-styles";
```
flycrum, line 1: I think this got converted to double quotes somehow. If you'd revert, that'd be appreciated.

coulterpeterson:
@flycrum just finished rebase and implementation of those changes. I must admit I've never had to do a merge-rollback-rebase before, so if anything looks awry please let me know 😊 I used the interactive rebase function of Fork and everything looks correct, but a second pair of eyes is always appreciated

PR #116, opened 2020-06-02
fix(LOC-1852): updated CSS hierarchy to match new page structure.

ethan309:
**Audience**

Local users.

**Summary**

Update to CSS for main site info page, needed to match new page structure (link length/color fix).


ethan309:
Ah, yes, I see now that I've missed those light mode users. Thanks, I'll make that fix right away!

ethan309:
For sure, I will take a look at using this instead. Thanks for the info!

ethan309:
The color check has been updated to match what is used in ` _theme.scss`.

PR #117, opened 2020-06-04
fix(LOC-1852): CSS restructuring for site overview layout.

ethan309:
**Audience**

Local users.

**Summary**

Update to CSS for site info overview page; this is needed to match new page structure (link length/color fix).

### DIFF
--- a/src/styles/containers/SiteInfo.scss
+++ b/src/styles/containers/SiteInfo.scss
@@ -6,6 +6,7 @@
 		display: flex;
 		flex-direction: column;
 		position: relative;
+		overflow: hidden;
 	}
 
 	.DownloaderOverlay {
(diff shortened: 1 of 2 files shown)


src/styles/containers/SiteInfo.scss
```diff
@@ -33,75 +34,87 @@
 		}
 	}
 
-	.SiteInfo_Top {
-		padding: 30px;
+	.SiteInfo_Header {
 		position: relative;
```
claygriffiths, line 38: Curious, why was `position: relative` required here?

src/styles/containers/SiteInfo.scss
```diff
@@ -33,75 +34,87 @@
...
+
+		.SiteInfo_Top {
+			padding: 30px;
+			position: relative;
+			width: fit-content;
+			max-width: 100%;
+
+			.SiteInfo_SiteLinkArea {
+				margin-right: 100px;
+				a {
```
claygriffiths, line 50: Nit-pick: missing newline after the `margin-right`. We need to implement SASS/SCSS linting. 😃

src/styles/containers/SiteInfo.scss
ethan309: Normally it would not make a difference. It only comes into play if an error banner pops up above the header section. Without a container with a relative position, the Start/Stop button will remain fixed and overlap the banner.

PR #120, opened 2020-06-08
fix(LOC-1905): correction to site info page so live link popups display properly

ethan309:
**Audience**

Local users.

**Summary**

Update the CSS for site info overview page; this fix allows live link popups to display properly without being cut off.

**Technical**

The `overflow` property of the overarching `SiteInfo` class was removed in favor of `min-width: 0`. This restricts the flexbox content from extending beyond the window size, but still allows "overflow" from elements such as the live link popups. [An extended explanation can be found here.](https://makandracards.com/makandra/66994-css-flex-and-min-width)

### DIFF
--- a/src/styles/containers/SiteInfo.scss
+++ b/src/styles/containers/SiteInfo.scss
@@ -6,7 +6,7 @@
 		display: flex;
 		flex-direction: column;
 		position: relative;
-		width: 0;
+		min-width: 0;
 	}
 
 	.DownloaderOverlay {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@getflywheel/local-components",
-	"version": "13.2.7",
+	"version": "13.2.8",
```
veryspry, line 3: No need to bump the version here! This will get taken care of in the deploy step. https://github.com/getflywheel/flywheel-local/wiki/Publishing-local-components

ethan309: Corrected, thanks for catching that.

PR #123, opened 2020-06-12
fix: resolve type issue in <TertiaryNav />

claygriffiths:
## Summary

`this.props.children` isn't a proper array so we need to get crafty with how to pull the first `<TertiaryNavItem />` so we can redirect to it.

### DIFF
--- a/src/components/menus/TertiaryNav/TertiaryNav.tsx
+++ b/src/components/menus/TertiaryNav/TertiaryNav.tsx
@@ -5,19 +5,34 @@ import * as styles from './TertiaryNav.sass';
 import IReactComponentProps from '../../../common/structures/IReactComponentProps';
 import { FunctionGeneric } from '../../../common/structures/Generics';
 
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@getflywheel/local-components",
-	"version": "13.3.3",
+	"version": "13.3.4",
```
veryspry, line 3: We've been holding off on version bumps in PR's and adding a separate commit for those at release time.

veryspry: https://github.com/getflywheel/flywheel-local/wiki/Publishing-local-components#bump-the-version-and-create-git-tag

claygriffiths:
@veryspry They do! https://reactjs.org/docs/react-api.html#reactchildrentoarray 

However, I ran into issues when trying to use it for this particular implementation so I stuck with mapping. I wish I would've jotted down the specific issue I ran into though. 🤔

PR #126, opened 2020-06-26
fix: yarn link issue

veryspry:
`yarn link`ing this package with `flywheel-local` has been causing TS errors due to differing `@types/react` versions in this package, this package's deps and Local. The fix is to use Yarn `resolutions` to ensure that the same version is used for all packages involved.


veryspry:
This thread is generally useful to understand the underlying issue
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25145

PR #135, opened 2020-08-18
Switch to Storybook and add GitHub Actions Workflow for Jest Tests

claygriffiths:
# Summary

* Switch from React Styleguidist back to Storybook
* Add GitHub Actions workflow that runs Jest and reports coverage to Code Climate
* Only build on `prepublishOnly` instead of `prepare` that way `local-components` doesn't build every time you `yarn`

**See individual commits for more details on the work that was done!**

# Context

This story started out as just adding a GitHub Actions Workflow for tests, but after digging in, the DevEx with `local-components` was severely lacking and Styleguidist would refuse to build in some cases. So, the migration to Storybook began!

# Technical

The vast majority of lines changed in this PR are `yarn.lock` and migrating from `README.md` files to `*.stories.mdx` files. For more details about the MDX format, see https://storybook.js.org/docs/react/api/mdx.

# Screenshots

_Behold, Storybook v6_

https://getflywheel.github.io/local-components has been updated as a part of the review process for this PR.

image


src/components/menus/TertiaryNav/examples/TertiaryNavWithDrawerExample.tsx
```diff
@@ -0,0 +1,90 @@
+/* @ts-nocheck */
```
veryspry, line 1: Nice! I love the inclusion of "recipe examples"

Does StoryBook expose these or are they just standalone for visual reference only?

tasks/create-component-entrypoints.ts
```diff
@@ -1,10 +1,10 @@
 /// <reference path="../src/global.d.ts" />
 import 'ignore-styles';
 
-import * as globby from 'globby';
-import * as path from 'path';
-import * as mock from 'mock-require';
-import * as fs from 'fs';
+import globby from 'globby';
```
veryspry, line 4: Nice!

src/components/menus/TertiaryNav/examples/TertiaryNavWithDrawerExample.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+class Item1 extends React.Component<IProps, { value: string }> {
+	constructor (props: IProps) {
+		super(props);
+
+		this.state = {
+			value: 'change this text',
+		};
+
+		this.handleChange = this.handleChange.bind(this);
```
veryspry, line 23: Not at all a showstopper in my book, but it would be nice to use arrow funcs and avoid calls to `bind`

PR #140, opened 2020-09-16
Star Component

alexasparks:
This PR adds an `IconCheckbox` component. The component accepts the following props: 

- `svgCheckedIcon`: React SVG component that accepts `className` prop, shown when input is checked
- `svgUncheckedIcon`: React SVG component that accepts `className` prop, shown when input is unchecked
- `onChange`: Function, passed the updated checked state as arg
- `checked`: Boolean, initial checked input state
- `disabled`: Boolean, disables input if true

Although called `IconCheckbox`, the purpose of this ticket was to create a Star Component for the Sites Organization work. To create the star component, just pass the new star svg icons to `IconCheckbox`. An example is in Storybook. This is what that component looks like:

**Checked w/ hover:** 

Screen Shot 2020-09-16 at 7 20 08 AM

**Unchecked w/ hover**

Screen Shot 2020-09-16 at 7 20 28 AM

**Checked w/o hover**
Screen Shot 2020-09-16 at 7 20 15 AM

**Unchecked w/o hover**
Screen Shot 2020-09-16 at 7 22 49 AM


claygriffiths (approved):
Love it. Thanks for going ahead and making this a generic component that we can reuse 🙌

Great screenshots in the PR, too!

src/components/inputs/IconCheckbox/IconCheckbox.scss
```diff
@@ -0,0 +1,27 @@
...
+.IconCheckbox {
+	height: 24px;
+	width: 24px;
+	position: relative;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+
+	&:hover {
+		background-color: #f7f6f6;
```
flycrum, line 11: Would you tweak this to use the sass variable `$gray5`?

flycrum: Now wondering if this is the value for both light and dark modes 🤔

flycrum: If so, perhaps you can reuse one of the mixins for that color background: 
image

flycrum (changes requested):
Really, really nice job on this story! I love the reusable component, great storybook examples, and excellent job including a test as well. Just a really minor color request/question and I think this one is ready for lift-on.

flycrum (approved):

PR #143, opened 2020-09-23
fix(LOC-2112): change dark mode hover color for IconCheckbox to work with any background

flycrum:
## Audience

Local Engineers and Add-on Developers

## Summary

IconCheckbox hover background color works on some backgrounds and not others for dark mode.

## Technical
- change hover for dark mode to darker gray (aka black) that we don't use for backgrounds so it always appears

### Hover
image

## Reference
- [LOC-2112](https://getflywheel.atlassian.net/browse/LOC-2112)


alexasparks (approved):
LGTM!

PR #144, opened 2020-09-23
fix(LOC-2112): fix IconCheckbox hover color and add alternative prop

flycrum:
## Audience

Local Engineers and Add-on Developers

## Summary

IconCheckbox hover background color works on some backgrounds and not others for dark mode.

## Technical
- change hover for dark mode to most commonly needed background color and add prop to use alternative for edge cases

## Reference
- [LOC-2112](https://getflywheel.atlassian.net/browse/LOC-2112)


alexasparks (approved):
LGTM! Thanks for taking care of this component!

PR #145, opened 2020-10-05
feat(LOC-2116): rewrite IconCheckbox to use 1 svg, allow color overrides, scale on active, and fix hover states

flycrum:
## Audience

Local engineers and Add-on developers

## Summary

`IconCheckbox` needs to override svg path fill colors by default and introduce props to toggle which states, if any, that it wants to use the original icon instead.

## Technical
- rewrite `IconCheckbox`to use 1 svg instead of 2 and dynamically set svg path fill colors for states
- remove circle hover bg because it wasn't consistent with the rest of the app and was conflicting with other bg styles
- add props for unchecked, unchecked hover, checked, and checked hover states to allow original svg colors to be applied
- add disabled style
- scale icon on mouse down (consistent with button effect)
- add fancy disable hover state to avoid weird hover state immediately following change of checkbox state
- add storybook examples for new props and goals

## Default styles with _all_ states overriding svg path fill colors (none of the original svg colors, by default)
Screen Recording 2020-10-05 at 12 09 38 PM

## Use star icon's original gold color for unchecked hover and checked states
Screen Recording 2020-10-05 at 12 13 24 PM

## Reference
[LOC-2116](https://getflywheel.atlassian.net/browse/LOC-2116)


alexasparks (approved):
LGTM

PR #149, opened 2020-10-29
style(LOC-2271): restyle site info menu item to remove upper-case and refactor to scss

flycrum:
# Summary:

Changing TertiaryNav from upper-case to default. Not only does this appear less aggressive and in your face, but it also provides better hierarchy and more horizontal space for labels, spinners, etc.

# Screenshots:

## New:
image

## Old:
image


src/components/menus/TertiaryNav/TertiaryNav.scss
```diff
@@ -0,0 +1,83 @@
...
+			display: inline-block;
+			width: 100%;
+			height: 36px;
+			line-height: 36px;
+			padding: 1px 20px 0;
+			border-radius: 4px;
+			color: $gray75;
+			font-weight: 500;
+			font-size: 14px;
+			@include tracking(40);
```
flycrum, line 44: Lost in moving this from .sass to .scss is the removal of this line which can more clearly be seen in [this commit](https://github.com/getflywheel/local-components/pull/149/commits/02fa75143861ee2269d91d92e138b12de450ba01):
```text-transform: uppercase;```

flycrum: This is the only change in this file btw.

tihal (approved):
LGTM! Thanks for migrating this to scss!

octalmage (approved):

PR #150, opened 2020-11-04
Rearrange ID's for input click tracking

alexasparks:
Move id to input elements for `IconCheckbox` & `InputPasswordToggle` components so global click tracking will recognize it as an 'input' for analytics.

Add an id to the option in `FlySelect` to track when a user clicks the option for analytics.

Accompanies: [PR for LOC-2307]( https://github.com/getflywheel/flywheel-local/pull/903)


veryspry (approved):
LGTM ⭐

PR #156, opened 2020-12-09
feat(LOC-2380): add ProgressRing component

flycrum:
## Summary

Create a new progress component that's circular (aka ring).

## Technical
- create new `ProgressRing` component within the `progress` section of storybook
- use programmatic svgs to efficiently render circle instead of outdated css hacks like 360 separate lines, clip-path, etc
- document with several story examples
- add basic tests
- flatten all CSS to make it easier to override when consuming
- add non-hashed class names (in addition to locally scoped names used for styling) for easier selector access later on
- clamp progress so it must always resolve to 0-100
- little tweaks to rotation of circle and progress fill math to make it "look full" at 100% while taking into account the extra size adding by `stroke-linecap: round`
- some of the colors used are automatically changed depending on theme light/dark when using default props values
- `mix-blend-mode` used for dark mode to get background circle colors just right for both dark gray and orange background colors

## Screenshots (as seen in Storybook)

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getflywheel/local-components",
-  "version": "15.2.0",
+  "version": "15.3.0",
   "repository": "https://github.com/getflywheel/local-components",
   "homepage": "https://build.localbyflywheel.com",
   "description": "",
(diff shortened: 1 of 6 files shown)


andrewbotz (approved):
Looks good. Had to get nit-picky to find something to change.

PR #160, opened 2021-02-01
feat: create Icons system and add Storybook option to show theme modes side-by-side

flycrum:
#### New Icon System

Icons were previously all one-off svgs that had a build script to convert them to .tsx components. As Local has scaled, this has resulted in several challenges:
* using these icons from other repos wasn't realistic and resulted in duplicate svgs across the repos
* discoverability of icons was challenging and required going file by file and previewing
* every icon was doing their own thing (some set explicit widths and heights, others stripped certain svg tags, others left raw data, every icon had their own color and intent, etc and it was challenging to identify which icons were doing what and why
* for the icons that didn't have an explicit width and height, the icons would often scale up and appear very large before a flicker would occur and it'd scale down to its intended size

An official Icons system was needed to provide solutions to each of these problems. This PR attempts to do just that and provides a storybook example to easily search for icons and grab a JSX snippet on how to use it.

#### Toggle Theme Views

Local supports light and dark mode out of the box but building components that account for both can be challenging. Making this challenge even harder was the fact that one could easily forget about one or the other mode only to discover styling bugs later on. To solve for this, a custom Storybook preview and menu item has been added that allows for side-by-side, stacked, and individual light/dark mode views. It's kind of cool 😎 

### Technical
* there is an `Icons-README.md` file with extensive details about objectives of the new system, how to add a new icon, q&a, and other specifics (check it out)
* there are two ways to import icons: 1) individually named imports and 2) The `Icons` namespace (the preview JSX can be toggled in the storybook example showcasing both usages)
* create `withIconSvg` hoc to streamline Icon component implementation, ensure consistency between Icons, allow neutralizing fill color, and easily providing metadata for the storybook example search criteria to make it more powerful and usable (especially as the library grows)
* add polish to icons story so it scrolls more logically as the list grows (header and footer should always be in view)
* support additional props for icons such as `size` (this can be expanded on later)
* * show snippet preview that factors in additional props (e.g. size) as well as import approach (which can be toggled)
* sync up selection of icons story example between light and dark modes using localStorage
* sync up search terms of icons story example between light and dark modes using localStorage
* clip to copy JSX snippet
* remove unused svgs from former one-off system
* `@babel/plugin-transform-typescript` added to support namespaces like `Icons`
* `storybook-addon-themes` needed to be updated to `6.x` to take advantage of new features that allowed for side-by-side and stacked views of light and dark modes
* several tweaks made to the storybook `preview.js` and `preview-head.html` templates

### Out of Scope (including some nice-to-haves)
* there are a lot of redundant icon svg files between `local-components` and `flywheel-local` and this story does not take on the additional step of moving the Local app to use this new system
* the old svgs still exist as every precaution regressions was taken...especially with the `neutralFill` option
* could break-up and componentize IconsStoriesAll more
* how to handle larger icons? scale for story? show scale %?
* show icon dimensions in preview
* how to manage colors?
* show no results and hide footer
* sort by relevance

#### The new split screen light/dark mode theme option

image

#### Icons now has its own dedicated category

image

#### Click to copy

image

#### The Icons search story with meta support

https://user-images.githubusercontent.com/41925404/106504555-b59b8780-648c-11eb-8450-163c305fe6ac.mov


src/components/icons/IconsStoriesAll.tsx
```diff
@@ -0,0 +1,209 @@
...
+						const Icon = getIconComponentRef(iconPropKey);
+						const exportNamespaceIconName = getExportNamespaceIconName(iconPropKey);
+						const meta: IconSvgStoryMeta | undefined = Icon?.meta;
+						const IconRef: React.FC = Icon as React.FC;
+
+						// filter out non-matching icons if there is search criteria
+						if (searchInputValue
+							&& !iconPropKey.toLowerCase().includes(searchInputValue.toLowerCase())
+							&& !exportNamespaceIconName?.toLowerCase().includes(searchInputValue.toLowerCase())
+							&& !meta?.tags?.join('`').toLowerCase().includes(searchInputValue.toLowerCase())
```
veryspry, line 180: What do you think about something like?
```suggestion
!meta?.tags?.map(tag => tag.toLowerCase()).includes(searchInputValue.toLowerCase())
```

src/components/icons/IconsStoriesAllFooter.scss
```diff
@@ -0,0 +1,111 @@
...
+.IconsStoriesAll_Footer {
+	@include theme-background-gray5-else-graydarkalt;
+	@include theme-border-top;
+	box-shadow: 0 -1px 5px 0 rgba(0, 0, 0, 0.04);
+
+	.IconsStoriesAll_AdditionalProps {
+		display: flex;
+		flex-wrap: wrap;
+		align-content: baseline; // eliminate space between rows
+		//padding-bottom: 10px;
```
veryspry, line 12:
```suggestion
```

src/components/icons/helpers/withIconSvg.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * HOC that adds additional props and styles to icons.
+ * @param IconComponent The icon component to be decorated by the hoc.
+ * @param neutralFill Whether to set the color of the svg path fill to the theme's neutral gray color.
+ * @param meta The meta details for an svg that makes an icon more discoverable within storybook example.
+ */
+const withIconSvg = <P extends object>(
+	IconComponent: any,
+	neutralFill: boolean,
```
veryspry, line 83: It looks like the majority of the calls to this function are getting `true` as their second arg. Thoughts on flipping the order of `neutralFill` and `meta` and then making the default value of `neutralFill` `true`?

Something like

```js
const withIconSvg = <P extends object>(
	IconComponent: any,
	meta: IconSvgStoryMeta,
	neutralFill: boolean = true,
): React.FC<P & IconSvgProps> => {
// ...
}
```

src/components/icons/svgs/CircleIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+		<circle cx="4" cy="4" r="4"/>
+	</svg>
+);
+
+export default withIconSvg(
+	Icon,
+	true,
+	{
+		tags: [
+			'round',
```
veryspry, line 20: Thoughts on adding `circle` to the list?

src/components/icons/svgs/CloseIcon.tsx
```diff
@@ -0,0 +1,64 @@
...
+			{
+				propName: 'size',
+				type: 'select',
+				options: ['s', 'l'], // todo: how to automatically get these options???
+				default: Icon.defaultProps.size,
+			},
+		],
+		tags: [
+			'dismiss',
+			'x',
```
veryspry, line 61: Thoughts on adding `close` to the list?

src/components/icons/IconsStoriesAll.tsx
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	function getExportNamespaceIconName(iconPropKey?: string): string | undefined {
+		if (!iconPropKey) {
+			return;
+		}
+
+		const Icon = getIconComponentRef(iconPropKey);
+
+		return iconsNamespacePropKeys.find((key) => (IconsNamespace as any)[key] === Icon) as string;
```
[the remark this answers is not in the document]
veryspry: There is no "action item" here, just my hot take :sweat_smile:

veryspry (changes requested):
This is an INSANE AMOUNT OF WORK. Thanks for tackling it; this is going to make all of our jobs so much easier and our  repos DRYer! :smile:

I'm "requesting changes" because of a few discussion points, but TBH, I don't think I'd want to die on any of those hills.

src/components/icons/IconsStoriesAll.tsx
```diff
@@ -0,0 +1,209 @@
...
+	set_doShowAdditionalProps: () => {},
+	doShowAdditionalProps: true,
+});
+
+const localStorageIconPropKey = 'IconsStoryAll-selected-iconPropKey';
+const localStorageSearchText = 'IconsStoryAll-selected-searchText';
+
+export const IconsStoriesAll = () => {
+	const contextValue = useContext(IconStoriesAllContext);
+	const [searchInputValue, set_searchInputValue] = useState('');
```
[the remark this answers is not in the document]
flycrum: 100% get that and it's not consistent with our rules. I'll make that change. The only shame is that it's so much easier to refactor names like this. But I totally agree with the mixed snake/camel case not being my cup-of-tea either and it's not something we're doing anywhere else.

src/components/icons/IconsStoriesAll.tsx
```diff
@@ -0,0 +1,209 @@
...
+	const [selectedIconData, set_selectedIconData] = useState<IconStoriesAllContextType['selectedIconData'] | undefined>(contextValue.selectedIconData);
+	const [doShowAdditionalProps, set_doShowAdditionalProps] = useState(contextValue.doShowAdditionalProps);
+	const [additionalPropChanges, set_additionalPropChanges] = useState<AdditionalPropChanges>(contextValue.additionalPropChanges);
+	const changedAdditionalProps: AdditionalPropChanges = {};
+	const iconsPropKeys: PropertyKey[] = Reflect.ownKeys(Icons); // use reflection to get all available icons
```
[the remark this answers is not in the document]
flycrum: Yeah, super dope especially since it's only for the example and doesn't make its way into our deployed package 😃

src/components/icons/helpers/withIconSvg.tsx
flycrum: It's a great thought. I actually did this initially to make a developer understand how they're implementing every given SVG and deciding whether a neutral fill actually makes sense. My goal is to work with Tyson and figure out a better way to do this and ultimately introduce a color prop for all icon components that makes this go away and automatically neutralizes all of them. Or, even better yet, let's get already neutralized assets and get rid of the ugly business of doing that behind-the-scenes. Let me talk to Tyson and figure out what we want to do here.

src/components/icons/svgs/CircleIcon.tsx
flycrum: Yeah, something I didn't document was that search checks the filename, import name, and these tags. But I like the idea of tags including everything, so this was a miss on my part. Thanks for catching it! I'll make the change.

src/components/icons/svgs/CloseIcon.tsx
flycrum: 💯

src/components/icons/svgs/CircleIcon.tsx
veryspry: Aha! That makes sense. In that case, I'm okay either way here!

src/components/icons/IconsStoriesAll.tsx
```diff
@@ -0,0 +1,209 @@
...
+	const iconsNamespacePropKeys: PropertyKey[] = Reflect.ownKeys(IconsNamespace);
+
+	function getIconComponentRef(iconPropKey?: string): IconType | undefined {
+		if (!iconPropKey) {
+			return;
+		}
+
+		return (Icons as any)[iconPropKey];
```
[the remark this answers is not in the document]
veryspry: Now that I'm re-reading this, my suggestion seems like one of those really Javascripty things. That said, I'm actually okay either way here.

src/components/icons/svgs/CircleIcon.tsx
flycrum: You know, I really like the idea of doing it for two reasons:

1. more explicit
2. if the icon is renamed at any point, for any reason, this keyword is still captured

Thanks for the feedback!

src/components/icons/IconsStoriesAll.tsx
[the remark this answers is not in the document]
octalmage: I'm ok with it in the context of tests, and a story is sort of like a test?

src/components/icons/Icons-README.md
```diff
@@ -0,0 +1,63 @@
+# Icons
```
octalmage, line 1: useful link to preview the rendered markdown: 

https://github.com/getflywheel/local-components/blob/079fcafdeae6f7538943a27c85aa1fd8fd738349/src/components/icons/Icons-README.md

src/components/icons/Icons-README.md
```diff
@@ -0,0 +1,63 @@
...
+    1. note: this will automatically include icons in the searchable storybook example via reflection
+
+## ⁉️ Q&A
+
+### Why does the svg size flicker at an absurdly large size before scaling down to a reasonable size?
+
+Chances are, the exported svg xml does not include an explicit `width` and `height` attribute
+but a developer can add this after the fact.
+
+### Why use <svg>? Why not <img>?
```
octalmage, line 35: LOL these tags need to be escaped some how. They don't render in the markdown.

src/components/icons/Icons-README.md
```diff
@@ -0,0 +1,63 @@
...
+1. the option to neutralize svg colors out of the box for consistency and predictability
+1. ability to override icon colors with a single `className` selector without using `!important` or nested selectors
+1. allow each icon to define their own metadata for ease of discovery and implementation
+1. dual export system that allows for importing via single namespace or individual component references
+1. icons should allow additional, unknown props to be passed through and applied
+
+## 🏗 Adding a new svg icon
+
+1. create a new icon component within `./src/components/icons/svgs/`
+1. copy the new svg markup to the rendered component
```
octalmage, line 18: What's the new SVG markup?

src/components/icons/Icons-README.md
```diff
@@ -0,0 +1,63 @@
...
+## ⁉️ Q&A
+
+### Why does the svg size flicker at an absurdly large size before scaling down to a reasonable size?
+
+Chances are, the exported svg xml does not include an explicit `width` and `height` attribute
+but a developer can add this after the fact.
+
+### Why use <svg>? Why not <img>?
+
+The problem with both <img> and background-images is that you don’t get to control the innards of the svg.
```
octalmage, line 37: This tag doesn't render either. maybe just backticks like ``

src/components/icons/Icons-README.md
flycrum: very helpful, thanks!

src/components/icons/Icons-README.md
flycrum: I thought magic might just take care of that 😅 Consider it done!

src/components/icons/helpers/withIconSvg.tsx
flycrum: I talked with Tyson and now have a plan to bake these values into the svg and that'll come in a soon-to-follow release. Thanks for the feedback!

src/components/icons/IconsStoriesAll.tsx
flycrum: Just tried it and it doesn't catch all search tags like the original does 🤔 I'll have to think through this again to see what it's doing and whether I can write something a little more intuitive as to what's going on and why.

src/components/icons/IconsStoriesAll.tsx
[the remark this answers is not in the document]
flycrum: Yeah, what made this particularly tricky was the reflection stuff. I'm sure there's a more graceful way of doing much of it and I'll try to find those, but the good news is just that Stallings, their effectively tests.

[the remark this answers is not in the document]
flycrum: Dang, this stuff is tricky! The combo of namespaces, keys as symbols, reflection, etc is rough to gracefully resolve

src/components/icons/IconsStoriesAll.tsx
veryspry: Totally! I'm cool with it as is and we can also iterate and improve in the future! :rocket:

veryspry (approved):
LGTM!

PR #161, opened 2021-02-19
feat(LOC-2671): create new Accordion component

flycrum:
# Summary

Create an Accordion component with:
- open/close action
- animation
- auto close any other items (optionally)
- default chevron
- stories
- tests

# Technical
- created 2 new components `Accordion` which is made up of a group of `AccordionItem` components.
- the `Accordion` component simply controls the state of each `AccordionItem` and auto-closes others if `allowMultipleOpenItems` is set to `false` (default)
- the `AccordionItem` does all the heavy lifting and can be used independent of the parent `Accordion` component
- a mixed control approach was used for `AccordionItem` that allows it to be dumb (controlled) or manage its own state (uncontrolled) and these two modes are auto-determined based on whether the `opened` prop is utilized or not
- both components can set the prop `clickArea` to determine whether the chevron (`icon`) is the lone button that opens/closes the content or if the entire summary area is clickable
- these components are keyboard accessible as this will continue to be a point of emphasis 🎉
- both components utilize the `Container` wrapper (which is now standard for most new components)
- added `react-animate-height` package to once and for all have non-janky height animation that's true to the actual content's height (all non-JavaScript approaches have well known limitations)
- lots of stories to cover a bunch of use cases
- basic tests
- added global storybook style `Story__Padding` that can be used on any story to apply a more aesthetically pleasing padding around story examples
- enhance standard SVG icon component props to allow for height and width overrides (if needed, rare)

# Screenshots

## Accordion stories

image

## AccordionItem stories

image

image

## Accordion Accessibility demo

https://user-images.githubusercontent.com/41925404/108534610-f45c7a80-729f-11eb-894f-03cdd9894218.mov

## AccordionItem Accessibility demo

https://user-images.githubusercontent.com/41925404/108534867-3f768d80-72a0-11eb-84b0-4a4ed6a7864b.mov

### DIFF
--- a/src/components/layout/Accordion/Accordion.scss
+++ b/src/components/layout/Accordion/Accordion.scss
@@ -1,0 +1,3 @@
+@import '../../../styles/_partials/index';
+
+.Accordion {}
--- a/src/components/layout/Accordion/Accordion.scss
+++ b/src/components/layout/Accordion/Accordion.scss
@@ -1,3 +1,0 @@
-@import '../../../styles/_partials/index';
-
-.Accordion {}
--- a/src/components/layout/Accordion/Accordion.test.tsx
+++ b/src/components/layout/Accordion/Accordion.test.tsx
@@ -1,0 +1,15 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import Accordion from './Accordion';
+import TestComponentPropUtils from '../../../utils/TestComponentPropUtils';
+
+describe('Accordion', () => {
+	it('renders without crashing', () => {
+		shallow(<Accordion />);
+	});
+
+	it('renders basic react props like id, className, and style as element attributes', () => {
+		const shallowWrapper = shallow(<Accordion {...TestComponentPropUtils.basicReactProps} />);
+		TestComponentPropUtils.expectsBasicReactProps(shallowWrapper, true);
+	});
+});
--- a/src/components/layout/Accordion/Accordion.test.tsx
+++ b/src/components/layout/Accordion/Accordion.test.tsx
@@ -5,11 +5,11 @@ import TestComponentPropUtils from '../../../utils/TestComponentPropUtils';
 
 describe('Accordion', () => {
 	it('renders without crashing', () => {
-		shallow(<Accordion />);
+		shallow(<Accordion><div /></Accordion>);
 	});
 
 	it('renders basic react props like id, className, and style as element attributes', () => {
-		const shallowWrapper = shallow(<Accordion {...TestComponentPropUtils.basicReactProps} />);
+		const shallowWrapper = shallow(<Accordion {...TestComponentPropUtils.basicReactProps}><div /></Accordion>);
 		TestComponentPropUtils.expectsBasicReactProps(shallowWrapper, true);
 	});
 });
--- a/src/components/layout/Accordion/AccordionItem.tsx
+++ b/src/components/layout/Accordion/AccordionItem.tsx
@@ -1,0 +1,184 @@
+import * as React from 'react';
+import classnames from 'classnames';
+import * as styles from './AccordionItem.scss';
+import { Container } from '../../modules/Container/Container';
+import ILocalContainerProps from '../../../common/structures/ILocalContainerProps';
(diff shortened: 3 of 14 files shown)


src/components/layout/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import Accordion from './Accordion';
+import TestComponentPropUtils from '../../../utils/TestComponentPropUtils';
+
+describe('Accordion', () => {
+	it('renders without crashing', () => {
```
veryspry, line 7: :joy:

src/components/layout/Accordion/Accordion.scss
```diff
@@ -0,0 +1,3 @@
+@import '../../../styles/_partials/index';
+
+.Accordion {}
```
veryspry, line 3: Thoughts on removing this file?

src/components/layout/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+function instanceOfPropsControlled(object: any): object is PropsControlled {
+    return 'opened' in object;
+}
+
+/**
+ * This is a factory of sorts that determines whether to use the controlled or uncontrolled variation.
+ * Uses controlled-specific prop(s) to determine whether the parent is trying to "control" state or not.
+ */
+const AccordionItem : React.FC<PropsControlled | PropsUncontrolled> = (props) => {
```
veryspry, line 172: Very cool!

src/components/layout/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,184 @@
...
+	return instanceOfPropsControlled(props) ? (
+		<AccordionItemControlled
+			{...props}
+		/>
+	) : (
+		<AccordionItemUncontrolled
+			{...(props)}
```
veryspry, line 179:
```suggestion
			{...props}
```

src/components/layout/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,184 @@
...
+					styles.AccordionItem,
+					'AccordionItem',
+					className,
+				)}
+				id={id}
+				style={style}
+			>
+				<>
+					{/* need to unwrap children because of potential extra 'child' wrapper (plus this works with fragments) */}
+					{ React.Children.map(children?.props?.children || children, (child: React.ReactNode, index: number) => {
```
veryspry, line 74: Wouldn't this be equivalent?
```suggestion
					{ React.Children.map(children, (child: React.ReactNode, index: number) => {
```

src/components/layout/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,184 @@
...
+												onClick={onClick}
+												type="button"
+											>
+												{!opened && icon === true && (
+													<CaretIcon />
+												)}
+												{opened && icon === true && (
+													<CheckmarkMixedIcon width={14} height={3} />
+												)}
+												{typeof icon === "object" && (
```
veryspry, line 110: Nice! Love the ease of overriding the icon here.

veryspry (approved):
LGTM!

src/components/layout/Accordion/Accordion.scss
flycrum: Yeah, I wrestled with this. I kept on thinking we might get some design feedback on how it scrolls and such, but that never came. I'll remove this. Thanks!

src/components/layout/Accordion/AccordionItem.tsx
flycrum: Oh ha! I'm not even sure how this happened 😅 Consider it changed.

src/components/layout/Accordion/AccordionItem.tsx
flycrum: Apparently not. I debugged this and given the unwrapping of the the intended parent element (if any), this actually does something different.

PR #165, opened 2021-04-01
fix: styling issue in banner component if no icon is present

tihal:
## Summary
This tiny PR bumps the local-components version and fixes an issue with Banners that removes an empty div from rendering if no icon is specified.

## Screenshots:

BEFORE: notice how the banner text is indented and not aligned left with the rest of the UI

Screen Shot 2021-04-01 at 1 28 52 PM

AFTER: fixed!

Screen Shot 2021-04-01 at 1 29 40 PM

## Reference

Fixed in process of working through LOCAL-194: https://getflywheel.atlassian.net/browse/LOCAL-194

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getflywheel/local-components",
-  "version": "15.7.1",
+  "version": "15.7.2",
   "repository": "https://github.com/getflywheel/local-components",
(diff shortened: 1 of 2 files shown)


flycrum (approved):
Nice fix! This was seriously annoying.

PR #169, opened 2021-04-09
refactor(LOC-1972) Rework FlyDropdown to use Tooltip and underlying React Portals

flycrum:
## Summary

FlyDropdown has the same cutoff issue that Tooltip used to and it is even worse within VirtualTable because of the unique way rows are laid out and cycled through in a virtual manner.

## Technical

* rather than reinventing the wheel and trying to implement React Portals into existing FlyDropdown code, use Tooltip to leverage its usage of Portals as well as styles, logic, lifecycle, etc
* add `popperOptions` to FlyDropdown to allow position tweaks for the underlying popup and arrow
* added `useClickInsteadOfHover` prop to repurpose Tooltip to be used for non-hover opening/closing
* converted FlyDropdown from class-based to functional component (initially because I was going to reinvent the wheel and use popper's hook)
* convert FlyDropdown's stylesheet from .sass to .scss (it never ends)
* update popper to use latest features (including padding positioning which is needed for FlyDropdown arrow)
* added missing standard `id`, `className`, and `style` props to Tooltip
* (unrelated) added static class names to VirtualTable to more easily overwrite its styles outside of local-components
* fix FlyTooltip which sneakily used `@extend` and broke when the FlyDropdown styles it was extending changed [commit](https://github.com/getflywheel/local-components/pull/169/commits/0a668410fd65999f43845698fc209c526c41937d)
* no longer hard set FlyDropdown width to `160xp` but rather use `min-width` for this [commit](https://github.com/getflywheel/local-components/pull/169/commits/6901a20bbc2eca35e4131590a7e47c8290e94c38)
* allow VirtualTable row height to be set to any pixel value rather than just predefined ones [commit](https://github.com/getflywheel/local-components/pull/169/commits/317caf19d2ba86f86d8bade723420f50d2f9916e)
* add prop `extraData` to VirtualTable in the event that the row or cell renderers need non table forming `data` info [commit](https://github.com/getflywheel/local-components/pull/169/commits/40c8457197bdc94b3cae80b7412e418fd631dbae)

### previous issue with VirtualTable

image

### fixed

image

## References
* [Popper API: Arrow Padding](https://popper.js.org/docs/v2/modifiers/offset/)


src/components/modules/VirtualTable/VirtualTable.tsx
```diff
@@ -73,8 +83,8 @@ export interface IVirtualTableProps extends IReactComponentProps {
 	onChangeRowData: (dataArgs: IVirtualTableOnChangeRowDataArgs) => void;
 	/** class to be applied to every row */
 	rowClassName: string;
-	/** Row height (using size values). */
-	rowHeightSize?: 's' | 'm' | 'l';
+	/** Row height as size values or explicit number of pixels */
+	rowHeightSize?: 's' | 'm' | 'l' | number;
```
tihal, line 87: Love this addition!

src/components/modules/VirtualTable/VirtualTable.tsx
```diff
@@ -8,13 +8,18 @@ import { VirtualTableHelper } from './helpers/VirtualTableHelper';
...
+}
+
 export interface IVirtualTableCellRendererDataArgs {
 	cellData: any;
 	changeFn: IVirtualTableProps['cellRendererChangeFn'];
 	/** the default rendered cell contents that can either be rendered within a custom cell renderer or ignored all together */
 	children: React.ReactNode;
 	colKey: string | number;
 	data: IVirtualTableContext['data'] | any;
+	extraData?: any;
```
tihal, line 22: Adding in extraData adds a whole new dimension of functionality here, thank you

tihal (approved):
Crum and I went over this in person, and I tested locally as well. Everything seems to be in order!

PR #170, opened 2021-04-26
Prepend componentWillReceiveProps with UNSAFE_

alexasparks:
As part of the effort to remove warnings from the dev console I renamed `componentWillReceiveProps` with UNSAFE_


flycrum (approved):
Thanks for cleaning this and those warnings! Eventually, we'll move these to functional components but this helps bridge the gap 🙌

PR #172, opened 2021-05-03
Warning icon not displayed properly when trying to push WP 5.4.2 version

alexasparks:
## The Problem

The warning icon was not displaying properly when trying to push changes to Flywheel with an incompatible WP version.

## Technical

Remove top margin style from `FlyModal.sass` that was preventing the `topIcon` from displaying right. 

## Screenshot

Before: 
image-20210426-191451

After: 
Screen Shot 2021-05-03 at 11 46 23 AM

### DIFF
--- a/src/components/overlays/FlyModal/FlyModal.sass
+++ b/src/components/overlays/FlyModal/FlyModal.sass
@@ -29,9 +29,6 @@
 		max-height: 93vh
 		padding: 60px
 
-		> div
-			margin-top: 0 !important
-
 	&.FlyModal__HasIcon
 		overflow: visible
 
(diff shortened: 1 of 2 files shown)


octalmage (approved):
I wish we had snapshot tests to know if this would break any other usages of the modal.

alexasparks:
> I wish we had snapshot tests to know if this would break any other usages of the modal.

I did do some spot testing with other modals to try and make sure, but agreed 💯 that snapshot tests for this would be better.

flycrum (approved):
Good fix! I'm sure something adjacent to this changed awhile back and broke this in the meantime.

PR #176, opened 2021-05-17
feat: add height prop to Banner

alexasparks:
Add a height prop to Banner.


flycrum (approved):
Hello banner height variations 🙌 LGTM

PR #178, opened 2021-06-04
feat: update BasicInput component to accept standard input attributes as props

tihal:
## Summary

Quick PR that adds the basic input attributes as props to the BasicInput component. Pretty straightforward!

Reference doc: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text

Also bumps local-components version.

All the props added here are forwarded directly to the underlying "input" component, so no additional work is needed aside from adding the props directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getflywheel/local-components",
-  "version": "16.1.0",
+  "version": "16.1.1",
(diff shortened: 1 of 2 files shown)


flycrum (approved):
Great improvement to a component that needs some love ❤️ Would you just add a quick note to the PR description that mentions that all of these props are forwarded to the underlying `input` component?

PR #190, opened 2021-09-10
fix(LOC-2801): refactor tooltip to correctly react to useClickInsteadOfHover prop

adamperrry:
We were conditionally using some hooks in the tooltip component based on the `useClickInsteadOfHover` prop, which was causing an error whenever that prop changed. This situation came up when modifying the FlyDropDown component.

This PR merges the two custom hooks into one and moves the condition inside. Note the element parameters were consolidated into 1 - this does remove the case where `useClickInsteadOfHover` was true but `null`s were passed to "bypass" - however, the dropdowns still work as expected with the right amount of event listeners, so not sure why that was being done.


src/components/overlays/Tooltip/Tooltip.tsx
```diff
@@ -245,12 +244,9 @@ const useTooltip = ({
 	const [ targetElement, setReferenceElement ] = useState<HTMLElement | null>(null);
 	const [ popperElement, setPopperElement ] = useState<HTMLElement | null>(null);
 	const [ transitionElement, setTransitionElement ] = useState<HTMLElement | null>(null);
-	const isHoverTarget = useClickInsteadOfHover
-		? useDetectClickWithinTargets(targetElement, null, true)
-		: useDetectHover(targetElement);
-	const isHoverPopper = useClickInsteadOfHover
-		? useDetectClickWithinTargets(null, null) // pass nulls to bypass otherwise this will conflict with other click detects (above)
-		: useDetectHover(popperElement);
+	const isHoverTarget = useDetectClickOrHoverWithinTargets({targetEl: targetElement, alwaysBlurOnClick: true, useClickInsteadOfHover})
```
alexasparks, line 247: What do you think about calling this `targetElement` and then you could just use: 

```suggestion
	const isHoverTarget = useDetectClickOrHoverWithinTargets({targetElement, alwaysBlurOnClick: true, useClickInsteadOfHover})
```

alexasparks: You'd have to change it below and in the hook too though

adamperrry: Hmm the use of popperElement on the next line makes it seem less worth it

alexasparks: Yeah, but it's kind of weird to have `targetEl` & `targetElement`. Maybe rename `targetElement`?

alexasparks (approved):

PR #191, opened 2021-09-15
feat(LOC-2997): add CopyInput component

adamperrry:
This PR adds a `CopyInput` component to Local-Components. It leverages the existing `CopyButton` component, with additions to that component to account for a new design. 

This component takes several props, and can show a label above the input field and a message below the input field. 

The component takes an `onChange` prop that is called on input change, passing the new value as an argument. This way, validation can handled by the parent component via the `onChange` and `invalid` props. The message can also be optionally shown only when invalid, making it an error message/CTA in that case.

There is also a `value` prop for manually setting the input value via the parent - while intended as an initial input value, any changes to this prop will update the value of the input field, not just initially. This may be useful for more complex validation behavior (a character limit, for example).

Finally, disabling the input via the `disabled` prop will remove the borders and align the text right, to be closer to the copy button. I did think on how to go about animating that text alignment, but since input field widths are not friendly at all, it wouldn't be easy. One option may be moving all of the input styling (borders, padding) to the wrapping div, utilizing `:focus-within` for the focus state. Then on disable, we could hide the input so it still takes up space, pop an identical, absolutely positioned text field on top, then slide that over. Too much work for now... but it could work!


src/common/helpers/useDetectClickOrHoverWithinTargets.ts
```diff
@@ -43,7 +43,7 @@ export const useDetectClickOrHoverWithinTargets = ({
 					// depending on whether to detect every click as cancelling something (e.g. tooltip)
 					// this should either toogle or set to true
 					setIsClickFocus((is) => alwaysBlurOnClick ? !is : true);
-					console.log("Inside onClick event")
```
alexasparks, line 46: Who let this slip through code review? 🤦‍♀️

src/components/buttons/CopyButton/CopyButton.tsx
```diff
@@ -35,6 +41,8 @@ export interface ICopyButtonProps extends ILocalContainerProps {
 	uncopiedStateIconVariant?: UncopiedStateIconVariants | keyof typeof UncopiedStateIconVariants | null;
 	/* Icon to show in copied state */
 	copiedStateIconVariant?: CopiedStateIconVariants | keyof typeof CopiedStateIconVariants | null;
+	/* Background styling for uncopied state */
+	uncopiedStateBGStyleVariant?: UncopiedStateBGStyleVariants | keyof typeof UncopiedStateBGStyleVariants;
```
alexasparks, line 45: I know you were just following a pattern here, but do you think it's necessary to have `UncopiedStateBGStyleVariants` in addition to the `keyof typeof UncopiedStateBGStyleVariants` ? Seems like just the `keyof typeof UncopiedStateBGStyleVariants` would be enough... unless I'm missing something?

src/components/buttons/CopyButton/CopyButton.tsx
```diff
@@ -62,34 +71,63 @@ export const CopyButton = (props: ICopyButtonProps) => {
...
+		copy(textToCopy);
+		if (isCopied) return;
+
+		const newTimeout = copiedTimeout + 2000;
+		const animationDuration = 475;
+
+		setPopout(true);
+		setTimeout(() => { 
+			setIsCopied(true);
+			setPopinPopout(true, false);
```
alexasparks, line 101: Looks like the value for `popin` & `popout` are always opposite, so I'm wondering if we can combine this into a shared variable? 

Something like: 
```
const handleClick = () => {
		copy(textToCopy);
		if (isCopied) return;

		const newTimeout = copiedTimeout + 2000;
		const animationDuration = 500;

		setTimeout(() => {
			setIsCopied(true);
			setPopin(true);
		}, animationDuration);
		setTimeout(() => setPopin(false), newTimeout - animationDuration)
		setTimeout(() => {
			setIsCopied(false);
			setPopin(true);
		}, newTimeout);
		setTimeout(() => setPopin(false), newTimeout + animationDuration);
	}
```

Then for the styles, what do you think of getting rid of the `popout` class & keyframes altogether and instead utilizing the `alternate` animation-direction? Something like: 

```
@keyframes popin {
	0% { transform: scale(0); opacity: 0; }
	40% { transform: scale(1.05); opacity: 1; }
	70% { transform: scale(.95); }
	100% { transform: scale(1); }
}

.popin {
	animation: popin .5s;
	animation-direction: alternate;
}
```

alexasparks: Mainly just trying to simplify this logic a bit

alexasparks: I'm just trying to think of ways to simplify this logic a bit 😃

src/components/buttons/CopyButton/_private/CopyButtonIcon.tsx
```diff
@@ -17,8 +19,16 @@ interface ICopyButtonIconProps {
...
 	const { variant } = props;
-	const { checkmark, checkmarkBackground } = CopiedStateIconVariants;
-	const { link } = UncopiedStateIconVariants;
+	const { checkmark, checkmarkBackground, checkmarkStroke } = CopiedStateIconVariants;
+	const { link, copy } = UncopiedStateIconVariants;
+
+	if (variant === copy) {
+		return (
+			<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+				<path fill-rule="evenodd" clip-rule="evenodd" d="M9 6C9.55228 6 10 6.44772 10 7V12V13C10 13.5523 9.55228 14 9 14H3C2.44772 14 2 13.5523 2 13V7C2 6.44772 2.44772 6 3 6H4H9ZM4 4V3C4 1.34315 5.34315 0 7 0H13C14.6569 0 16 1.34315 16 3V9C16 10.6569 14.6569 12 13 12H12V13C12 14.6569 10.6569 16 9 16H3C1.34315 16 0 14.6569 0 13V7C0 5.34315 1.34315 4 3 4H4ZM6 4V3C6 2.44772 6.44772 2 7 2H13C13.5523 2 14 2.44772 14 3V9C14 9.55228 13.5523 10 13 10H12V7C12 5.34315 10.6569 4 9 4H6Z" fill="#51BB7B"/>
```
alexasparks, line 28: I think `fill-rule` & `clip-rule` need to be camelcase, I'm seeing some errors in the console about it.

src/components/buttons/CopyButton/CopyButton.tsx
adamperrry: Don't think you're missing anything, pretty sure we only want `keyof typeof`, though below a prop is being passed `as CopiedStateIconVariants`, but removing the union doesn't throw an error, so might as well

src/components/buttons/CopyButton/CopyButton.tsx
adamperrry: Ugh yeah I hit a refactoring wall at this point - just toggling the class won't work because the animation has a single cycle that's triggered only when the class is applied. In my mind you should still be able to take the class off and immediately reapply it using a single variable, but `alternate` really only seems to apply when you have more than one cycle in the animation - like the next time you reapply the class it won't be in reverse.

I did try using just one `keyframes` declaration, defining the animation in the `popout` class to use those keyframes in reverse, but even that doesn't work. The `keyframes` seem to only want to be used by one thing at a time. So separating everything out very explicitly is the only way it likes to behave :/

src/components/buttons/CopyButton/_private/CopyButtonIcon.tsx
adamperrry: Aha good catch, I just assumed that error wasn't me for some reason 🥴

src/components/buttons/CopyButton/CopyButton.tsx
```diff
@@ -62,34 +71,54 @@ export const CopyButton = (props: ICopyButtonProps) => {
...
+	}
 
-	const isTransparentCopiedStateBG = copiedStateBGStyleVariant === CopiedStateBGStyleVariants.transparent;
+	const handleClick = () => {
+		copy(textToCopy);
+		if (isCopied) return;
+
+		const newTimeout: number = copiedTimeout + animationDuration * 2;
+		setIsCopiedAndTriggerPopin(true);
+		setTimeout(() => setIsCopiedAndTriggerPopin(false), newTimeout);
```
alexasparks, line 89: I really dig this refactor 🔥 🔥

alexasparks (approved):

src/components/inputs/CopyInput/CopyInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+	/* Whether value is invalid - shows the invalid state of text input if true */
+	invalid?: boolean;
+	/* Label to be shown above text input */
+	label?: string;
+	/* Message to be shown underneath an enabled input e.g. length requirement, example input, etc. */
+	message?: string;
+	/* Only show message when invalid is true - useful for only displaying error/validity CTAs */
+	onlyShowMessageWhenInvalid?: boolean;
+	/* Options for CopyButton Tooltip */
+	copyButtonTooltipOptions?: TooltipProps;
```
alexasparks, line 18: Can you rename this `copyButtonTooltipProps` ? Just for clarity since that's what we'll be passing through?

src/components/inputs/CopyInput/examples/CopyInputExample.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
+import { CopyInput } from "../CopyInput";
+
+export const CopyInputExample = () => {
+	const [invalid, setInvalid] = React.useState(false);
+	const onChange = (newValue: string) => {
+        setInvalid(newValue.length < 5 ? true : false);
+	};
+        const offset: [number, number] = [0 , 15];
```
alexasparks, line 9: There's something weird with the formatting here... maybe a mix of tabs & spaces?

alexasparks (changes requested):

alexasparks (approved):

PR #192, opened 2021-09-16
feat(LOC-2998): updated drawer component

adamperrry:
For this ticket, we could leverage the existing `drawer` component, which basically had all the behavior we needed already! I converted the sass to scss and added two props for us to get to the desired design (remove stripes and add drop shadow). 

I also tweaked the bezier curve for the pop-in and pop-out animations to add a bit more pop. Colors approved by Tyson on the dark mode drop shadow, which I added to `_theme.scss`, with a new variable for the default rgba() shades of black. 

Video:

https://user-images.githubusercontent.com/62191853/133690617-950f1e60-123c-48b7-8424-3717b16e710a.mov


src/components/menus/Drawer/Drawer.tsx
```diff
@@ -1,19 +1,26 @@
...
+	shadow?: boolean;
 }
 
 interface IState {
 	disableAnimation: boolean;
 }
 
 export default class Drawer extends React.Component<IProps, IState> {
+	static defaultProps: Partial<IProps> = {
+		stripes: true,
```
alexasparks, line 20: Can we make the default `stripes: false`? Since the designs so far don't include stripes & that way the implementation of this component could be `<Drawer stripes />` instead of `<Drawer stripes={false} />` ?

adamperrry: I'm hesitant because defaulting to true is just defaulting to existing behavior, which I feel like is safest. I could actually change it to `noStripes`, that would preserve existing any existing implementation without refactoring and allow for the implicit prop passing - neat!

alexasparks (approved):

PR #194, opened 2021-09-17
feat: add hideTooltip prop to conditionally hide tooltip

alexasparks:
Add prop to conditionally hide tooltip so we don't have to keep duplicating code when using the `Tooltip` component.


adamperrry (approved):
Ship it ASAP! 🙌🏼

PR #196, opened 2021-09-23
feat(LOC-2981): changes for editing LiveLinks

adamperrry:
This PR contains a few changes to some components that were necessary while completing [LOC-2981](https://wpengine.atlassian.net/browse/LOC-2981). The ugliest change here is nesting the scss in `CopyInput.scss` to allow for overriding the `TableList` styles, which is possible due to the global scoping of those classes.

Otherwise the git commits describe the changes - allowing for a drawer component in `TertiaryNav`, setting `visibility: hidden` when not showing the drawer to prevent a tabbing bug, and, most frustratingly, adding a background color to the drawer component. I spent forever wondering why that component was sitting behind the content, when actually it just didn't have a background 🥴

### DIFF
--- a/src/components/inputs/CopyInput/CopyInput.scss
+++ b/src/components/inputs/CopyInput/CopyInput.scss
@@ -1,74 +1,75 @@
 @import '../../../styles/_partials/index.scss';
 @import '../../../common/styles/themeUtils';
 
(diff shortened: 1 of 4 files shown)


octalmage (approved):
Looks good! Great inline comments!

PR #197, opened 2021-09-23
feat(LOC-2981): CopyInput and Banner tweaks

adamperrry:
This PR tweaks the CopyInput to change the message size and change the "readonly" state from using the "readonly" attribute to the "disabled" attribute. 

Also allows for removing stripes on the banner component and using the large close button instead of the smaller close icon.


octalmage (approved):
I hope no one was using readonly!

PR #199, opened 2021-09-27
feat(LOC-2986): add optional tooltip to CopyInput input field

adamperrry:
This PR wraps the input field of `CopyInput` in an optional Tooltip and adds the `inputTooltipProps` prop, much like how the tooltip around the `CopyButton` component works.

This required applying much of the input CSS to the wrapping div, as it surrounds both the input field and the CopyButton component, which are now both surrounded by a ToolTipTargetContainer.

### DIFF
--- a/src/components/inputs/CopyInput/CopyInput.tsx
+++ b/src/components/inputs/CopyInput/CopyInput.tsx
@@ -74,7 +74,7 @@ export const CopyInput = (props: ICopyInputProps) => {
 			>
 				<Tooltip
 					{...inputTooltipProps}
-					hideTooltip={inputTooltipProps ? inputTooltipProps.hideTooltip : true}
+					hideTooltip={inputTooltipProps ? inputTooltipProps?.hideTooltip : true}
 				>
 					<input
 						className={styles.CopyInput__Input}
@@ -91,7 +91,7 @@ export const CopyInput = (props: ICopyInputProps) => {
 				<Tooltip 
 					className={styles.CopyInput__CopyButton} 
 					{...copyButtonTooltipProps}
-					hideTooltip={copyButtonTooltipProps ? copyButtonTooltipProps.hideTooltip : true}
+					hideTooltip={copyButtonTooltipProps ? copyButtonTooltipProps?.hideTooltip : true}
 				>
 					<CopyButton
 						copiedStateIconVariant="checkmarkStroke"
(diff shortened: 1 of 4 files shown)


src/components/inputs/CopyInput/CopyInput.tsx
```diff
@@ -66,26 +69,30 @@ export const CopyInput = (props: ICopyInputProps) => {
...
-					placeholder={placeholder}
-					type="text"
-					value={textToCopy}
-					disabled={disabled}
-					autoComplete="off"
-					spellCheck="false"
-				/>
+				<Tooltip
+					{...inputTooltipProps}
+					hideTooltip={inputTooltipProps ? inputTooltipProps.hideTooltip : true}
```
tihal, line 77: Could you change the condition on the ternary to be `inputTooltipProps?.hideTooltip`? That way we know for sure that value is defined.

tihal (changes requested):

src/components/inputs/CopyInput/CopyInput.tsx
```diff
@@ -74,7 +74,7 @@ export const CopyInput = (props: ICopyInputProps) => {
 			>
 				<Tooltip
 					{...inputTooltipProps}
-					hideTooltip={inputTooltipProps ? inputTooltipProps.hideTooltip : true}
+					hideTooltip={inputTooltipProps ? inputTooltipProps?.hideTooltip : true}
```
tihal, line 77: Ah, my bad! I meant the first instance, like so:

`hideTooltip={inputTooltipProps?.hideTooltip ? inputTooltipProps?.hideTooltip : true}`

tihal (changes requested):

src/components/inputs/CopyInput/CopyInput.tsx
adamperrry: Oh yeah I see, my change did nothing haha. But I think that changes the intent a little - I want the tooltip to default to hidden when no TooltipProps are passed, and default to showing when they are passed but `hideTooltip` isn't. I think if `hideTooltip` isn't defined, it'll be falsey

adamperrry: Could change it to:

`hideTooltip={inputTooltipProps ? (inputTooltipProps.hideTooltip ?? false) : true}`

but I think that's redundant since undefined is falsey

tihal: Ah right! With undefined resolving to falsey then we should be good. Approving!

tihal (approved):
✅

PR #204, opened 2021-10-15
fix: pass local checked state to change handler

alexasparks:
Before v16.5.2 the `Switch` component passed the local state checked value to the `onChange` handler. This pr reverts back to that behavior. I also removed the `checked` value from the destructured object and just accessed that value with `props.checked` to make it immediately apparent that there were 2 checked values 1 from props & 1 in local state.

### DIFF
--- a/src/components/inputs/Switch/Switch.tsx
+++ b/src/components/inputs/Switch/Switch.tsx
@@ -38,10 +38,13 @@ const Switch = (props: IProps) => {
 	React.useEffect(() => setChecked(props.checked), [props.checked])
 
 	const handleChange = () => {
-		setChecked(!checked);
-		if (onChange) {
-			onChange(name, !checked);
-		}
+		setChecked((prev) => {
+			if (onChange) {
+				onChange(name, !prev);
+			}
+
+			return !prev;
+		});
 	}
 
 	return (


src/components/inputs/Switch/Switch.tsx
```diff
@@ -34,14 +33,14 @@ const Switch = (props: IProps) => {
...
 
-	const [isChecked, setIsChecked] = React.useState(checked)
+	const [checked, setChecked] = React.useState(props.checked)
 
-	React.useEffect(() => setIsChecked(checked), [checked])
+	React.useEffect(() => setChecked(props.checked), [props.checked])
 
 	const handleChange = () => {
-		setIsChecked(prev => !prev);
+		setChecked(!checked);
```
adamperrry, line 41: Could we keep this as `prev => !prev`? I think it's best practice to use the callback syntax when you need to set a state that depends on the previous state value, since these are asynchronous calls that don't happen immediately

src/components/inputs/Switch/Switch.tsx
```diff
@@ -34,14 +33,14 @@ const Switch = (props: IProps) => {
...
 		if (onChange) {
-			onChange(name, checked);
+			onChange(name, !checked);
```
adamperrry, line 43: Good catch, sorry about that! Hope that wasn't a huge headache but I bet it was

src/components/inputs/Switch/Switch.tsx
adamperrry: You actually could probably do something like: 
```
setIsChecked((prev) => {
    if (onChange) onChange(name, !prev);
    return !prev;
})
```
to guarantee the onChange handler gets the right value

src/components/inputs/Switch/Switch.tsx
alexasparks: Lol, no worries. I just thought I was going crazy for a minute when things stopped working in production 😆

src/components/inputs/Switch/Switch.tsx
alexasparks: I dig it

PR #210, opened 2021-11-10
feat: add yield icon to LocalComponents

adamperrry:
Adds Yield Icon to local components, to be used in Outage Banner:

image


tihal (approved):
LGTM!

PR #211, opened 2021-11-18
feat: updates to BasicInput and InputPasswordToggle

adamperrry:
This PR adds error states/messages to both BasicInput and InputPasswordToggle. It also converts InputPasswordToggle to a functional component and adds a variation for using a labeled checkbox for showing the password instead of the Eye SVG. 

The InputPasswordToggle doesn't seem to have been touched since the repo's inception, so I also cleaned up the styles a little by pulling the global styles into a scoped module. 

image

image


tihal (approved):
Looking good! 

Nice cleanup, and always satisfying to convert old components to functional components.

PR #212, opened 2021-11-18
feat: add Alert component

adamperrry:
This PR adds an Alert component per Tyson's design (to be initially used in the Connect Modal redesign). The alert is optionally dismissible via props, and the optional `show` and `onDismiss` props allow for even greater control over dismissing/showing the alert.

image


src/components/alerts/Alert/Alert.tsx
```diff
@@ -0,0 +1,74 @@
...
+		container,
+		children,
+		...otherProps
+	} = props;
+
+	const [showAlert, setShowAlert] = useState(show);
+
+	const handleDismiss = () => {
+		setShowAlert(false);
+		if (onDismiss) onDismiss();
```
tihal, line 38: I'm rusty on my component lifecycle, but will this cause console errors to the effect of "attempting to call function within an unmounted component"?

tihal (commented):
Just one Q before approval!

src/components/alerts/Alert/Alert.tsx
adamperrry: I don't think so - tested it in story book with a console.log, and it registers

tihal: Sweet!

tihal (approved):
Good to go!

tihal (approved):
Approved!

PR #231, opened 2022-03-03
feat(LOC-3301): add ContextMenu and testing-library/react

adamperrry:
This PR adds a new 3-dot `ContextMenu` component that borrows heavily from `FlyDropdown`. The work also required some changes in `Tooltip` and `ButtonBase`, but those are just extensions. We had some discussions about what belongs in Local Components and what the interface should be, and I've kind of landed on exposing most of the tooltip interface in `ContextMenu` for now, as most if not all of that interface is visual stuff, except for `content`. I've also exposed `onHide` and `onShow` to hook into the menu, perhaps for tracking.

The `items` prop expects a type pretty similar to that expected by React's context menu (`MenuItem`), which is why I went the route I did there with a lot of the optional properties. In core we should be able to encapsulate the data we pass to that context menu and send it to both the original one and this new component. 

Additionally, I added `testing-library/react` and a few related packages - [this library](https://testing-library.com/docs/react-testing-library/intro) really knocks it out of the park and is a much better testing framework than Enzyme. I added a test for `ContextMenu` to show what it can do as far as relevant tests that a user would care about that don't dive into implementation details. 

This addition also required an upgrade to `jest` and `ts-jest`, so the `yarn.lock` changes look intimidating, but it's a really powerful library. 

https://user-images.githubusercontent.com/62191853/157302248-e53c79d0-460b-4f1c-b9d0-a6dbab9dcf20.mov


src/components/buttons/_private/ButtonBase/ButtonBase.tsx
```diff
@@ -140,20 +143,24 @@ const ButtonBase = (props: IButtonBaseProps) => {
...
 	const handleKeyDown = (e: React.KeyboardEvent) => {
-		if (e.key === 'Enter') {
+		if (e.key === 'Enter' && !active) {
 			setIsActive(true);
 		}
 	}
 
 	const handleKeyUp = (e: React.KeyboardEvent) => {
-		if (e.key === 'Enter') {
+		if (e.key === 'Enter' && !active) {
```
austin-redick, line 155: Should this check be `if (e.key === 'Enter' && active) { ... }` since you follow with setting `active` to false afterwards?

bgturner (commented):
We had a great conversation in Around related where components should live as well as how much testing should be done. I feel good about next steps. Adam is looking into ways of limiting the interface of the Tooltip

src/components/buttons/_private/ButtonBase/ButtonBase.tsx
adamperrry: I follow with setting `isActive` to false. `active` is a prop for forcing the button to be active, like `disabled`. This code is basically saying only let the enter key add or remove the `_active` class if the parent isn't specifying that the button should be active.

src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { TextButton } from '../../buttons/TextButton/TextButton';
+
+export interface IMenuItem {
+    color?: 'red' | 'none';
+    className?: string;
+    content?: React.ReactNode;
+    label?: string;
+    onClick?: FunctionGeneric;
+    enabled?: boolean;
+    type?: 'normal' | 'separator';
```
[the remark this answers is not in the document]
adamperrry: Also, added a check to filter out empty menu items passed to `ContextMenu`

src/components/buttons/_private/ButtonBase/ButtonBase.tsx
austin-redick: I'm following now. I got `active` and `isActive` mixed up I think.

.editorconfig
```diff
@@ -1,9 +0,0 @@
-root = true
-
-[*]
-charset = utf-8
-end_of_line = lf
-indent_style = tab
-insert_final_newline = true
-trim_trailing_whitespace = true
-indent_size = 4
```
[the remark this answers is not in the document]
adamperrry: I have some code in place to get linting set up with eslint and prettier, I'll make a separate ticket/PR for that and add back the editor config file at that point

src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export interface IContextMenuProps extends IReactComponentProps {
+    /** className for the list items' container */
+    classNameList?: string;
+    /** className for an individual list item */
+    classNameListItem?: string;
+    /** whether to force the tooltip to show and ignore mouse events **/
+    forceHover?: boolean;
```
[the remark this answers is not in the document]
adamperrry: I looked into what I'd want to limit from the tooltip props, and really all of them are just very innocuous visual adjustments that I'd hate to take out just to wish we had them in the future. The one thing I did remove was `content`, which will be based on `items`. This one specifically I like being able to control the open state via a prop, could be good for a future walkthrough for example

src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+    /** array of menu items */
+    items: IMenuItem[];
+    popperOptions?: TooltipProps;
+    position?: 'top' | 'bottom';
+    useClickInsteadOfHover?: boolean;
```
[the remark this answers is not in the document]
adamperrry: For this I'm picturing somebody using this as maybe a "More" menu that activates on hover? I don't think it's the worst prop to have, especially once I write some tooltip tests.

src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+const ContextMenu = (props: IContextMenuProps) => {
+    const {
+        className,
+        classNameList,
+        classNameListItem,
+        items,
+        popperOptions,
+        position,
+        useClickInsteadOfHover,
+        ...otherProps
```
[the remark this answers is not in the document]
adamperrry: Yeah this is pretty common in the repo, eslint-react by default does not like it though since it says it makes it unclear which props are being sent to the component... I'd have to disagree with them there

tihal:
I'm a simple man, I see 1700 lines changed and I think "wow that's impressive, give Adam a raise"

src/common/helpers/LaunchPlayground.ts
```diff
@@ -0,0 +1,22 @@
...
+// to copy and paste it like an animal. Only tested on macOS. Use and enjoy but
+// maybe don't commit any call sites.
+//
+// Add `import { LaunchPlayground } from 'common/helpers/LaunchPlayground';`
+// to the top of your file or test and call like so:
+// `LaunchPlayground(window.document.body);`
+// element: a chunk of the DOM or a query result from the `screen` object in the
+// testing harness.
+//
+// returns happiness
```
alexasparks, line 15: Lol, love this

src/components/buttons/Button/Button.tsx
```diff
@@ -34,11 +34,12 @@ export const Button = (props: IProps) => {
...
-			color={setIntentColor(props, ButtonPropColor.gray)}
+			color={setIntentColor(props, ButtonPropColor.green)}
 			id={id}
-			form={setForm(props, ButtonPropForm.outline)}
-			padding={ButtonPropPadding.m}
-			style={style}
+			secondary
+			form={setForm(props, ButtonPropForm.fill)}
+			padding={ButtonPropPadding.s}
+			style={{ lineHeight: 1.5, ...style }}
```
alexasparks, line 42: Curious, why not apply this `lineHeight` style to the 'Button' class instead of passing it as part of the style object?

src/components/inputs/FlyDropdown/FlyDropdown.tsx
```diff
@@ -24,7 +24,7 @@ interface IProps extends IReactComponentProps {
 	/** className for an individual list item */
 	classNameListItem?: string;
 	/** whether to force the tooltip to show and ignore mouse events **/
-	forceHover?: boolean;
+	forceShow?: boolean;
```
alexasparks, line 27: Dig this prop name change!

src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+    };
+
+    const [isShowing, setIsShowing] = React.useState(false);
+
+    const content = (
+        <ul className={classnames(styles.ContextMenu_Items, 'ContextMenu_Items', classNameList)}>
+            {items
+                .filter((item: IMenuItem) => Object.keys(item).length !== 0)
+                .map((item: IMenuItem, i: number) => (
+                    <li key={i} className={classnames(styles.ContextMenu_Item, classNameListItem, item.className)}>
```
alexasparks, line 44: What is the difference here between `classNameListItem` and `item.className`? Do we need both?

src/components/overlays/FlyTooltip/FlyTooltip.scss
```diff
@@ -6,7 +6,7 @@
 	display: inline-block;
 
 	&:hover,
-	&.FlyTooltip_Container__ForceHoverState {
+	&.FlyTooltip_Container__forceShowState {
```
alexasparks, line 9: This is such a nit, but can you capitalize the 'f' for the class?

src/components/overlays/Tooltip/Tooltip.tsx
```diff
@@ -320,13 +323,18 @@ export const Tooltip = (props: TooltipProps) => {
...
 	});
 
-	const isShowing = (forceHover || !stages.isStage0Hidden) && !hideTooltip;
+	const isShowing = (forceShow || !stages.isStage0Hidden) && !hideTooltip;
+	const isFirstRender = useRef(true);
 
 	useEffect(() => {
-		if (isShowing) {
-			onShow?.();
+		if (isFirstRender.current) {
```
alexasparks, line 330: Curious about why you need this?

alexasparks (changes requested):

src/components/buttons/Button/Button.tsx
adamperrry: `Button` is a global style, I try to avoid adding any more global styles. The `Button` and `PrimaryButton` components don't have CSS files since they're just basically configurations of `ButtonBase`, and I don't want to put it in ButtonBase, because then `TextButtons` will get it - it's really these two buttons that will have consistent font sizes.

src/components/overlays/Tooltip/Tooltip.tsx
adamperrry: On first render, `onHide()` is called without the check, just an annoying React quirk. See https://stackoverflow.com/questions/53253940/make-react-useeffect-hook-not-run-on-initial-render

src/components/menus/ContextMenu/ContextMenu.tsx
adamperrry: So `classNameListItem` are styles that will be applied to every `MenuItem`, and `item.className` can be used to target a specific `MenuItem`

src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+        item.onClick?.call(null);
+        event.stopPropagation();
+    };
+
+    const [isShowing, setIsShowing] = React.useState(false);
+
+    const content = (
+        <ul className={classnames(styles.ContextMenu_Items, 'ContextMenu_Items', classNameList)}>
+            {items
+                .filter((item: IMenuItem) => Object.keys(item).length !== 0)
```
austin-redick, line 42: Just a thought in case there are simpler use cases in the future, but instead of chaining filter and map here I think you could achieve the same with one reduce to save some iterations. It could get tricky with the types though and I like this for its readability.

src/components/overlays/FlyTooltip/FlyTooltip.scss
adamperrry: Done!

src/components/overlays/Tooltip/Tooltip.test.tsx
```diff
@@ -1,9 +1,71 @@
...
+const triggerText = 'Tooltip trigger';
+const trigger = <div>{triggerText}</div>;
+
+const renderTooltip = (props?: TooltipProps) => {
+    return render(<>
+		<Tooltip position="right" content={content} showDelay={0} hideDelay={0} {...props}>
+			{trigger}
+		</Tooltip>
+	</>);
+};
```
austin-redick, line 21: Really like what you did with the tooltip tests and testing-library in general!

austin-redick (approved):
The new ContextMenu is cool!

alexasparks (approved):

PR #232, opened 2022-03-11
feat: update colors - mostly grays - for contrast compliance

adamperrry:
This PR searches for `$gray25` and `$gray75` text in the app and in most cases replaces it with `gray` in light mode. I also made some small tweaks to other components to pass accessibility standards. In some cases I lightened the gray color used in dark mode as well.


dani-j-d (approved):
Nice! Code changes look good and appearance appears to reflect correct colors.

PR #233, opened 2022-03-11
feat: add external link icon

adamperrry:
Adds the external link icon to Local Components, previously only lived in Core.


dani-j-d (approved):
Good to go! Nicely done!

PR #234, opened 2022-03-12
fix: fix tooltip trigger on click

adamperrry:
This PR fixes a bug where clicking the trigger element for a click-trigger tooltip opens the tooltip, but clicking inside the tooltip and clicking the trigger again doesn't close it until the second click. 

The tooltip component uses a custom function to set up hooks for the trigger and popper elements to determine whether to close or open the popper: `useDetectClickOrHoverWithinTargets`. When `useClickInsteadOfHover` is `true`, there are extra listeners and handlers to determine when to to close the popper using a state variable called `isClickFocus`.

6 months ago I consolidated the two separate functions into just this one, and introduced a bug where clicking the open popper messes up the state for `isClickFocus` for the trigger element. This PR adds another optional argument to the function called `ignoreClickOn`, where an element can be passed on which clicks will be ignored when setting that state and removing event listeners.

### DIFF
--- a/src/common/helpers/useDetectClickOrHoverWithinTargets.ts
+++ b/src/common/helpers/useDetectClickOrHoverWithinTargets.ts
@@ -3,11 +3,13 @@ import { useEffect, useState } from 'react';
 export const useDetectClickOrHoverWithinTargets = ({
 	targetEl, 
 	alwaysBlurOnClick, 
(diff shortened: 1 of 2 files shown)


dani-j-d (approved):
Code and functionality check out. Nice solve!

PR #235, opened 2022-03-23
feat: review updates to buttons, contextMenu, tooltip, and stylesheet precedence

adamperrry:
This PR adds a few things:
- Update to Tooltip to `focus()` the popper container on open, meant for tabbing accessibility (note the focus isn't trapped)
- Re-adds `.editorconfig` (linting PR to come, but that won't require a version bump thankfully)
- Some tweaks to buttons again, reverting some previous tweaks and updating the secondary button/outline color scheme per new designs
- Added ClockIcon
- Adjustments to ContextMenu per design, just colors
- Change to `add-styles` to attach the stylesheet to the top of `head`, not the bottom. This fixes an issue where when developing with linked Local-Components in Local core, every time the linked package rebuilds (either manually or in watch mode), add-styles is run and the linked package takes precedence over the app styles (in `app.css`). Ideally this wouldn't be an issue, but it does affect a few styles in core that are probably using `!important` or something - regardless, that stylesheet isn't added in production in that location, so this is just better matching production behavior by giving it lower precedence. This will affect where add-on stylesheets are loaded, but those are scoped to the specific Local Components version, so it shouldn't affect them at all.


src/add-styles.ts
```diff
@@ -46,6 +46,6 @@ export default function () {
 	link.type = 'text/css';
 	link.rel = 'stylesheet';
 
-	document.getElementsByTagName( 'head' )[0].appendChild( link );
-
+	const head = document.getElementsByTagName( 'head' )[0];
+	head.insertBefore( link, head.firstChild );
```
alexasparks, line 50: Great find - so excited for this change!

src/components/icons/svgs/ClockIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import Clock from '../../../svg/clock.svg';
+import withIconSvg from '../helpers/withIconSvg';
+
+export const ClockIcon: React.FC = ({...props}) => {
```
alexasparks, line 5: Why spread the props argument here (line 5)?

PR #236, opened 2022-03-23
feat: TextButtonExternal and further tweaks

adamperrry:
This PR moves TextButtonExternal from Local core to Local Components. 

It also creates a utility class for site labels, which in Local Core were often just using `__Color__Green`, which conflicted with the updated styles.

Finally, I made a tweak to the tooltip `focus()` mechanism, so that it only focuses once when showing. Before, we were focusing it on every change to the popper element, which was pulling focus away from inputs in modals that were popped by the `ContextMenu` component.

### DIFF
--- a/src/components/buttons/TextButtonExternal/TextButtonExternal.stories.mdx
+++ b/src/components/buttons/TextButtonExternal/TextButtonExternal.stories.mdx
@@ -25,7 +25,7 @@ Setting the `underline` prop to true means the link will always be underlined. T
 	</Story>
 </Canvas>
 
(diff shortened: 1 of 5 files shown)


bgturner (changes requested):
Nice stuff @adamperrry ! 

I notice that this element doesn't have an entry in Storybook. Can you create an entry so that non-devs can get a feel for this component and the various settings available to it?

bgturner (approved):
Looks good! I pushed two last minute quick-fixes, a typo and removed the `trackKey` prop

PR #238, opened 2022-03-29
feat: various updates per design review feedback

adamperrry:
This PR adds a few changes that came up during a call with Tyson where we discussed the UI refresh work. 

The commit messages are pretty descriptive of what was done!

### DIFF
--- a/src/components/menus/ContextMenu/ContextMenu.tsx
+++ b/src/components/menus/ContextMenu/ContextMenu.tsx
@@ -50,15 +50,14 @@ const ContextMenu = (props: IContextMenuProps) => {
                                 onClick={(event) => onClickItem(event, item)}
                                 disabled={item.enabled === undefined ? false : !item.enabled}
                                 intent={item.color === 'red' ? 'destructive' : 'default'}
-                                style={{ flexGrow: 1, padding: '5px 0', lineHeight: '20px' }}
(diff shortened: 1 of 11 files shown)


src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -50,15 +50,14 @@ const ContextMenu = (props: IContextMenuProps) => {
                                 onClick={(event) => onClickItem(event, item)}
                                 disabled={item.enabled === undefined ? false : !item.enabled}
                                 intent={item.color === 'red' ? 'destructive' : 'default'}
-                                style={{ flexGrow: 1, padding: '5px 0', lineHeight: '20px' }}
-                                privateOptions={{ fontWeight: 'medium' }}
+                                style={{ flexGrow: 1, padding: '5px 0' }}
```
dani-j-d, line 53: Thank you for this change... I had spotted this previously and was agitated LOL

dani-j-d (approved):
Nice! Code looks good and functionality is as expected

src/components/menus/ContextMenu/ContextMenu.tsx
adamperrry: Yeah, it was all of the sudden included in the designs, but Tyson said we should probably just remove it, thankfully!

PR #239, opened 2022-03-30
feat: update components per design review

adamperrry:
This PR updates the button colors and SVG margins per design feedback and the Local design system Figma doc.

It also adds a prop to ContextMenu to make the background option for a more lowkey dropdown menu.

Finally, we adjust a global style that was tageting the "Heads up" message when downloading a lightning service.


austin-redick (approved):
I like the conditional style classes!

PR #240, opened 2022-04-05
feat(LOC-3357): RadioBlock updates and linting setup updates

adamperrry:
This PR updates the linting configuration for Local Components. This also bumps the node dependency to 14.17.0 for devs. 

The biggest change is updates to RadioBlock component to allow us to complete the Blueprint work. Changes include:
- Update to Functional Component
- Fixes for conditional corner border radius
- CSS adjustments to move things to the Option div instead of Wrapper div
- `content` prop added to allow for content to be passed via each option instead of just a title. This can be used to add a description, button, or completely override the RadioBlock option. We can also now more easily pass class names and styles for finer control. For example: 

image


dani-j-d (approved):
Looks and works great! No nits, except that maybe "blueprint" related information should not be in the library. I would maybe use a different phrase or term since this is a generic and publicly available library. I understand the need to keep things moving so I am not gonna push on this. Approved!

PR #243, opened 2022-04-08
fix: add z-index to scrollshadow shadow div

adamperrry:
This PR adds a z-index to the scroll shadow's shadow div. Also linted the ScrollShadow component. 

I attempted to do a larger refactor because implementing the scrollshadow component is pretty complicated and requires a wrapping div with `position: relative; overflow: hidden` on it, and the children contents to be contained in a single wrapping div. 

I was going to pull those wrappers into the ScrollShadow component, but it was really messing with existing implementations, so I may just tackle that in the future.


dani-j-d (approved):
LGTM! The changes match what we discussed earlier and I don't have any nits with your code!

PR #244, opened 2022-04-08
feat: update Stepper component and grow steps

adamperrry:
This PR adds a `flex-grow: 1` to the `Step` component in `Stepper`. Before, we were calculating the width of the steps based on the number of steps, though I think this change will be much more flexible. Also, I think it still looks really good if the screen is small enough to start adapting to the size each individual step.

I also updated the component in general to a functional component using best practices. 

Also includes further tweaks to RadioBlock styling per Tyson's feedback.


dani-j-d (approved):
LGTM!

PR #247, opened 2022-04-12
feat: adjust radioblock container and add z-index to tooltip popper

adamperrry:
This PR adjusts the `RadioBlock` component once again! We were originally placing radio block options inside of an optional `Container` element, so I just moved the outer radio block option wrapper to that `Container` element. So now we will always render that `Container` (defaults to `div`), but we are still able to pass custom options to it. This makes adding a tooltip to a radio block option with content possible without style conflicts!

I also added a z-index to the tooltip popper container, because due to a bad bug fix for the user profile avatar showing up in front of overlayed window components, it was being hidden by a z-index of 10 on the `window` component. I created a ticket to tackle this problem more elegantly in the Popup component (use a tooltip or portal instead of a z-index), but this should work for now. 

I linted a couple files here too, which makes this look a little scarier.

### DIFF
--- a/src/common/structures/IReactComponentProps.ts
+++ b/src/common/structures/IReactComponentProps.ts
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 export default interface IReactComponentProps {
+	onKeyDown?: any;
 	children?: React.ReactNode;
 	className?: string;
 	key?: string | number;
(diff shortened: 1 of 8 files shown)


dani-j-d (approved):
Dude! This was a killer solve on such short notice! LGTM!

dani-j-d (approved):
LGTM!

PR #248, opened 2022-04-14
fix: make radioBlockItem key the option value instead of label

adamperrry:
This PR is the last one to RadioBlock I swear. Fixes a "unique key in list" console error when using options without labels.


bgturner (approved):
Woohoo!

PR #249, opened 2022-04-14
fix: increase z-index for FlyModal and more updates

adamperrry:
This PR fixes a bug that was caused by a bug fix in tooltip to a bug caused by a bad bug fix in popup to a bug caused by a z-index there. We've been bumping z-indexes and forgot this one :/ I also:

- Fixed a problem with the RadioBlock outline on hover with > 2 options
- Added a `selectedIcon` to `Flydropdown` for Cloud Backups - will just plop an icon to the left of the trigger TextButton, no real styles added
- Added a `disableStyle` to the `FlyDropdown` trigger `textButton`. Note it won't disable the whole Flydropdown, just make it look disabled.
- Refactired BasicInput to a function component and added `autofocus` capability with Refs!

### DIFF
--- a/src/components/buttons/_private/ButtonBase/ButtonBase.tsx
+++ b/src/components/buttons/_private/ButtonBase/ButtonBase.tsx
@@ -141,8 +141,8 @@ const ButtonBase = (props: IButtonBaseProps) => {
 	} = props;
 
 	const Tag: any = tag;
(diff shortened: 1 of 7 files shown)


dani-j-d (approved):
LGTM!

dani-j-d (approved):
Unless you have more changes... LGTM!

PR #253, opened 2022-05-05
feat: remove dtsBundleWebpack

adamperrry:
This PR removes `dts-bundle-webpack` (a wrapper around `dts-bundle`), because I don't think we actually need it to generate a merged typings file. Actually, the `typings.d.ts` file that it was generating wasn't being used by the IDE because the `typings` field wasn't being set in `package.json`, so we really wont' feel this change. 

This should resolve a critical dependabot on `minimist`, or at least open the door for it, since this package was not being maintained.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "homepage": "https://build.localbyflywheel.com",
   "description": "",
   "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


bgturner (approved):
lgtm!

PR #254, opened 2022-05-05
feat: upgrade url-parse and remove it from package json

adamperrry:
This PR removes the `url-parse` package from `package.json` (it's still installed as a dependency), and replaces it with the newer `URL` api available in browsers. Also includes a commit from dependabot to upgrade to a new minor version.


bgturner (approved):
lgtm!

PR #256, opened 2022-05-06
chore: upgrade storybook to 6.4.22 and other dependabots

adamperrry:
This PR upgrades `@storybook/react` from 6.1.11 to 6.4.22. Had a little trouble with this one since I had to add some Babel plugins - also couldn't figure out why the JSX in `preview.js` wasn't being interpreted correctly, but switching to a `.tsx` file fixed it right up. 

Should resolve an `immer` critical dependabot issue I believe? Might also resolve a few more dependabot issues actually.

Also threw in some commits from Dependabot: 
- `minimist` version bump 
- `moment` version bump
- `follow-redirects` version bump
- `tar` version bump


bgturner (approved):
lgtm

dani-j-d (approved):
Very nice! Appears to function as expected!

PR #259, opened 2022-05-11
chore: upgrade sb, bump react types, remove theme addon

adamperrry:
This PR fixes issues with `react-router-dom` typings by upgrading to the latest beta version of storybook. See [this github issues for more context](see https://github.com/storybookjs/storybook/issues/16837#issuecomment-984006866). To completely resolve that issue I also had to remove the `storybook-addon-themes` plugin, but I've been here almost a year and never used the button that added to switch between light and dark mode, so now it just defaults to the split view.

There was also an issue where a `react-testing-library` is depending on `@types/react^17`, which causes our hoisted module to be `@types/react@16.14.26`, causing a mismatch between versions between Local Components and Core. So I decided to just bump it to that hoisted version in both repos, and that resolved those build errors.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -7,7 +7,6 @@ module.exports = {
 		"../src/**/*.stories.mdx",
 	],
 	"addons": [
-		'storybook-addon-themes',
 		{
 			name: '@storybook/addon-essentials',
 			options: {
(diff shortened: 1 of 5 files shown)


dani-j-d (approved):
LGTM!

dani-j-d (approved):

PR #262, opened 2022-05-18
feat: various postcss related updates and cleanup

adamperrry:
Oof this PR took a lot of digging! Still not super satisfied in the end because I had to selectively upgrade packages to NOT their latest versions, but in the end I got everyone on the same version of PostCSS, and everything seems to work.

I removed both task files, since we don't actually need separate javascript files for each component as far as I can tell - `index.js` really should be enough. The other task file wasn't being run at all. I also deleted `scoped.sass`, last modified 3 years ago, probably a remnant. This was all really confusing because I had to learn how we've got our CSS solution set up. 

Basically:

- We need a `scoped.css` file and a `global.css` to exist in `dist`, because those are the stylesheets we add in Local.
- `scoped.css` is generated by webpack - it finds all imported css and packs it up through the loader chain and generates a single scoped css file (with CSS modules) via the `extract-css-chunks-webpack-plugin`, which itself is a little dated
- `global.sass` isn't included in that webpack bundle because it's not imported anywhere in the javascript, so we have to generate that manually
- We do that with `yarn build:sass` by first compiling the sass to css with `sass,` then parsing the CSS Modules syntax (`:global`) with `postcss-cli` configured to use the `postcss-modules` plugin (I removed the `postcss-url` plugin because we don't actually have any `url()` css syntax in this `global.css` file, and it drags down the `postcss` dependency version.
- I really wanted to just get rid of the `postcss-cli,` but I couldn't find a good way to parse that CSS Modules syntax. We can't just remove the `:global` syntax from the files referenced by `global.sass` because they import helper functions that then end up with the `:global` syntax themselves. 

In the end, I think this works for now. Oh and also, upgrading `style-loader` breaks Storybook for some reason.

## Other approaches?

I was thinking maybe we could get the global styles added to what we're now calling `scoped.scss` by importing `global.sass` in `index.ts` - that would put everything through the same loader chain, and maybe we wouldn't need `yarn build:sass` anymore. I haven't tried that yet though...

### Update
Just tried the above and I think that totally works. That's basically what storybook was doing in `preview.js`. So now I'm just gonna remove `postcss-cli` and `postcss-modules`, and we won't have a `global.css` file anymore - everything will just be in `scoped.css`. 

### To test:
Link up Local components, run `yarn build` or `yarn watch` and remove in Local core any imports of `dist/global.css`


bgturner (approved):
Looks good to me. To test, I:

1. linked components to core
2. ran the various `yarn` run tasks that were changed. Everything built cleanly, and style changes were reflected in Core
3. Removed `global.css` references in core to verify that styling seems the same and the browser console didn't have an error

PR #265, opened 2022-05-31
feat: add resolution for trim package

adamperrry:
This PR adds a resolution for the `trim` package to use `^0.0.3`, fixing a dependabot security issue. Storybook still working, which is why this package was here in the first place.


bgturner (approved):
checked this out locally and everything still seems to work!

PR #268, opened 2022-06-02
feat(LOC-4138): upgrade to webpack 5

adamperrry:
This PR upgrades from webpack 4 to 5 and also upgrades a whole host of plugins and loaders to their latest versions. The motivation here was to switch to `@svgr/webpack` for our svg loader to resolve an `nth-check` dependabot issue without a resolution. I also switched to the `mini-css-extract-plugin` for pulling our css into one file. 

This required some changes to the storybook config as well, including config for webpack 5 and a fallback polyfill for `os` in `untildify`, since node polyfills are no longer provided by default. I believe we didn't have to do that in the base webpack config because we have our target set to electron - either way, confirmed that the browseInput button component still works by linking everything up.

The one thing that was stopping me up for a long time was CSS modules, but I just had to update how we were importing the style objects: 
from 
```javascript
import * as styles from **/*.css
``` 
to 
```javascript
import styles from **/*.css
```


bgturner (approved):
Nice! I tried linking components with core. On save:
- Storybook updated correctly
- Core's `build.dev` process noticed the change, and the renderer showed the update after refresh.

PR #269, opened 2022-06-23
feat (LOC-3986): Build out new combobox component

dani-j-d:
AC:
- Component is built and implemented to match designs: https://www.figma.com/file/83UXFJyJBu2JFtgJT7mBWX/%F0%9F%92%9A-%F0%9F%9A%80-Local-Finals?node-id=8097%3A21812 --> Done

To test:
- pull and checkout branch
- run yarn build
- run yarn start
- open the browser and select the Storybook tab
- in Storybook, look under Inputs and select Combobox
- select the Connect case
- type, search, select from the dropdown, etc. to test against the designs (see design snapshot below)

Screen Shot 2022-06-23 at 10 23 23 PM


src/components/inputs/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import Combobox from './Combobox';
+import TestComponentPropUtils from '../../../utils/TestComponentPropUtils';
+
+describe('Combobox', () => {
+	// it('renders without crashing', () => {
```
austin-redick, line 7: Does this commented out section need to be re-written? Or, can it be removed?

src/components/inputs/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,564 @@
...
+		invalidMessage,
+		noResultsMessage = 'None found. Try a new search term.',
+	} = props;
+
+	const formatOption = (
+		option: ComboboxOption,
+		optionValue: ComboboxOptionFormatted['value']
+	): ComboboxOptionFormatted => {
+		return {
+			value: optionValue,
```
austin-redick, line 104: If you don't format this to add an explicit optionValue, where would the key value (that you are trying to guarantee) of state.options come from? I was having a little trouble following.

austin-redick (commented):
Mostly looks good and that Combobox component is a doozy! I just had some trouble following parts and had a few clarifying questions.

src/components/inputs/Combobox/Combobox.test.tsx
adamperrry: Deleted this test file for now, will probably add some tests later this afternoon or tomorrow

src/components/inputs/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,564 @@
...
+	| ComboboxOptionFormatted[]
+	| string[]
+	| { [value: string]: ComboboxOption }
+	| { [value: string]: string };
+export type ComboboxOptionsLoader = () => Promise<ComboboxOptions>;
+
+export interface ComboboxOptionGroup {
+	label: string;
+	linkText?: string;
+	linkOnClick?: FunctionGeneric;
```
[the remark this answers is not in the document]
adamperrry: This is coming from FlySelect, so that's a great question. I can maybe write a story for it in Storybook

src/components/inputs/Combobox/Combobox.tsx
adamperrry: Not sure I understand your question

src/common/helpers/useForwardedRef.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const useForwardedRef = <T>(ref: React.ForwardedRef<T>) => {
```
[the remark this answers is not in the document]
adamperrry: Yeah for sure! 

So `useForwardedRef` basically allows us to use a forwarded ref inside a component that receives a forward ref. By use, I mean act on, like focus the input in our case. Normally when you forward a ref, you just pass it right along and don't do anything but make it available to the parent component. But this hook basically merges two refs and keeps them in sync.

useCombinedRefs is kind of similar except it combines 2 refs and keeps them in sync. In our case I wanted to turn a ref we were already using into a forwarded ref available to any parents, so we need to merge the two to keep them in sync.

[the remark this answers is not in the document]
adamperrry: And needed to do this because we need refs to the basic input for height and position calcs on that DOM element, and you can only get a ref to a functional component if you forward the ref. Same reason we needed to do this in tooltip, because combobox can be disabled with a tooltip.

src/components/inputs/Combobox/Combobox.tsx
austin-redick: I'm basically wondering why it's necessary to have a formatted and unformatted version of the component. Are they mutually exclusive in their usage?

adamperrry: Ah, it's to allow a flexible interface for the options prop. Like you can pass an array, and object, or an already formatted object. Internally, the component logic only ever uses formatted options so that we can rely on the data structure.

PR #271, opened 2022-07-05
fix: scroll to option when navigating via keyboard

adamperrry:
We switched up the focus approach when navigating via keyboard in Combobox, resulting in keyboard navigation that focused/hovered options without scrolling to them in a scrollable dropdown. I added a simple solution of using a callback ref to get access to the DOM element on every render, scrolling to the option if it is focused. 

I also changed the default `maxHeight` to 0 instead of undefined to prevent a glitch on initial render that caused the dropdown to fully expand before getting the `maxHeight` applied.


dani-j-d (approved):
LGTM!

PR #273, opened 2022-07-18
fix(LOC-4290): overflow secondary text same as primary

adamperrry:
Small adjustments to make the secondary text a readable color and truncate properly.

Note the flex basis for these items is `auto`, so the width of the truncated text will depend on the length of the actual text itself. 

image


bgturner (approved):
This does deal with the overflowing text by truncating it with an elipsis, so I'm good with approving.

From the end-user's perspective, I can imagine it would be frustrating to not be able to see all of the text for the various sites that I'm trying to work with. In an ideal world, I think it would be good to explore a design solution that somehow allows a user to access the full text. 

My mind goes to some sort of text wrapping, but there may be other options out there. 

Anyway, long title's might be an edgecase we won't encounter often, so I'll be interested to see if we get any feedback from end-users.

alexasparks (approved):

PR #276, opened 2022-07-22
fix: reenable aria-checked on radioblock due to container component

adamperrry:
We need to make RadioBlock more accessible. In the meantime, we need it to be selectable by PlayWright, to easily understand whether an option is selected or not via selectors. 

This PR re-enables the `aria-checked` attribute, which was originally being passed until I changed the wrapper component to `Container`.

### DIFF
--- a/src/components/modules/Container/Container.tsx
+++ b/src/components/modules/Container/Container.tsx
@@ -58,7 +58,7 @@ export const Container = React.forwardRef((props: IContainerProps, ref: React.Fo
 			onKeyDown: props.onKeyDown,
 			tabIndex: props.tabIndex,
 			ref,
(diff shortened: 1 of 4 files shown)


src/components/modules/Container/Container.tsx
```diff
@@ -57,6 +58,7 @@ export const Container = React.forwardRef((props: IContainerProps, ref: React.Fo
 			onKeyDown: props.onKeyDown,
 			tabIndex: props.tabIndex,
 			ref,
+			'aira-checked': props['aria-checked'],
```
austin-redick, line 61: typo

austin-redick (approved):
LGTM

PR #278, opened 2022-09-14
feat(LOC-4310): banner redesign

adamperrry:
This PR includes a redesign of the banners. I pulled the state and style for the carousel out of the Banner component and into the BannerCarousel component. This will require changed implementation in core (where the BannerCarousel is not even currently used, but will need to be). 

I also reintroduced the themed preview for Storybook, as I incorrectly thought it was part of a plugin, but it's actually built in to Storybook now. I am defaulting to light mode instead of split, because some components behave weirdly when split (this one included, due to ID's being used to tie labels and hidden radio button inputs for accessibility).

Another note is the change to the DragBar height from 23px to 10px. This is because the banner carousel in core sits pretty high up, with the top padding on the first carousel radio button being 10px from the top - and drag bar was getting in the way.

[Link to storybook on Chromatic](https://62d18352969d7c650b935483-fyibwakjyz.chromatic.com/?path=/story/alerts-bannercarousel--banner-carousel-interactive)

image


src/components/buttons/Close/Close.tsx
```diff
@@ -9,10 +9,11 @@ import { ButtonBase } from '../_private/ButtonBase/ButtonBase';
 interface IProps extends IReactComponentProps {
 	position?: 'absolute' | 'static';
 	onClick?: FunctionGeneric;
+	size?: 's' | 'm' | 'l';
 }
 
 const Close: React.FC<IProps> = (props: IProps) => {
-	const { position = 'absolute', onClick, className, ...otherProps } = props;
+	const { position = 'absolute', size = 'l', onClick, className, ...otherProps } = props;
```
bgturner, line 16: This isn't something that will make me block this PR, but I'd love to start seeing changes in the props of our components be reflected in the corresponding storybook. So something like what's outlined in this doc:

https://storybook.js.org/docs/react/writing-stories/args

This would help us have a better living documentation.

src/components/icons/svgs/CircleInfoIcon.tsx
```diff
@@ -1,50 +1,49 @@
...
+Icon.defaultProps = {
+	size: 's',
+};
+
 export default withIconSvg(Icon, true, {
+	additionalProps: [
+		{
+			propName: 'size',
+			type: 'select',
+			options: ['s', 'l'], // todo: how to automatically get these options???
```
bgturner, line 44: What does this comment mean? Do you mean like how can we expose a large/small option for all icons?

bgturner (approved):
Looks good!

This PR was large, but it made more sense when I broke it down into two main parts:

1. New icons and the logic to normalize the rendering of them as well as better display of those icons in the "Icons" section of Storybook. 
2. Updates to the Banner Carousel

I like the updates that you're making to how the icons show in storybook, the main thing I'm seeing is the use of the `additionalProps` property which allows us to visually select the different style variations in Storybook.

Nice job normalizing the sizes (ie `small`/`sm`/etc ) into `s` `m` `l`.

Thanks for all the improvements!

src/components/alerts/BannerCarousel/BannerCarousel.tsx
```diff
@@ -1,39 +1,267 @@
...
+
+		return () => clearTimeout(activeTimeout);
+	}, []);
+
+	// useLayoutEffect to grab height of dummy banner BEFORE the browser paints
+	// allows us to set the height without visual glitches
+	useLayoutEffect(() => {
+		if (gettingHeight) {
+			// Carousel button height will equal 13px per unselected banner, plus 20px for selected, plus 20px padding
+			const buttonsHeight = numBanners > 1 ? (numBanners - 1) * 13 + 20 + 20 : 0;
```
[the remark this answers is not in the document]
adamperrry: Yeah this was to avoid using another `ref` for the carousel. Could separate out the numbers into named variables to make it clearer, but I'm inclined to keep the math at least since it works pretty well

PR #280, opened 2022-10-07
feat: change from flex grow to width 100

adamperrry:
I was using `flex-grow: 1` on the banner to fill the banner carousel, but we can feasibly have banners without the carousel (though probably shouldn't). In that case, say it's a flex container with a column direction, the flex grow would affect the height, not the width. I should've used `width: 100%` instead.


dani-j-d (approved):
Easy change, no negative impact. LGTM!

PR #281, opened 2022-10-10
fix: add z-index 100 to downloaderoverlay

adamperrry:
We are expanding the DownloaderOverlay in SiteInfo to cover the bottom bar, which has a popup element in the Live Links toggle that utilizes a z-index. 

This PR adds a z-index of 100 to the DownloaderOverlay, which I believe is one of the only reasonable use cases for z-index!


dani-j-d (approved):
Nice! LGTM!

PR #283, opened 2022-10-24
feat: use @electron/remote for BrowseInput

adamperrry:
This is the only blocker for the Electron upgrade in Local Components. 

BrowseInput needs to access the remote module to launch a file Explorer dialog. I added the dependency to `package.json` - previously, `electron` didn't need to be a dependency because that was the target for webpack. However, since this is a separate package, we need the dependency - I tried just ignoring it in Webpack via that IgnorePlugin, but linking it up didn't work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getflywheel/local-components",
-  "version": "17.6.1-beta.6",
+  "version": "17.6.1-beta.7",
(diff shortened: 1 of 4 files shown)


bgturner (approved):
Nice! Storybook compiles and opens successfully, and linking to Local Core continues to open Electron's dialog!

PR #284, opened 2022-10-28
feat: align table list row div child to middle

adamperrry:
Wow this is my first time touching these TableList styles and they are so convoluted and hard to follow. But adding this vertical align for the table-cell div children really makes things look a lot nicer, especially One Click Admin and the new Xdebug toggle.


dani-j-d (approved):
LGTM!

nickcernis (approved):

PR #288, opened 2022-11-18
feat: improve BannerCarousel to prevent extra rerenders

adamperrry:
The BannerCarousel is good, but it can be better!

This PR changes the carousel to use a `banners` prop instead of the `children` prop, so that we can see more clearly when the banners in the carousel have changed. This is easy because in Core we already have an array of this type that we modify in order to pass to BannerCarousel, so this will simplify things in core a little. 

I was previously just relying on a count, but sometimes the banners in a carousel will change without the number of banners changing. In that case, I had a stopgap to just update the children array, but when implemented in Core, we have a bunch of rerenders that would clear timeouts when switching between sites, for example. In that case the banner would just disappear while being hidden. 

With this change, the banner dismissal timeout will complete properly! I also added in some checks to allow us to pass raw JSX (type `React.ReactNode`) as the banner component or a function that returns JSX (like a render prop).

### DIFF
--- a/src/components/alerts/BannerCarousel/BannerCarouselExample.tsx
+++ b/src/components/alerts/BannerCarousel/BannerCarouselExample.tsx
@@ -6,7 +6,7 @@ import BannerCarousel from './BannerCarousel';
 
 interface IBanner {
 	id: string;
-	component: any;
+	component: React.ReactNode;
 }
 
 const loremIpsum =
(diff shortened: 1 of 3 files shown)


src/components/alerts/BannerCarousel/BannerCarouselExample.tsx
```diff
@@ -6,7 +6,7 @@ import BannerCarousel from './BannerCarousel';
 
 interface IBanner {
 	id: string;
-	component: any;
+	component: React.ReactNode;
```
dani-j-d, line 9: I love that you are leaning into stricter typing!

src/components/alerts/BannerCarousel/BannerCarouselExample.tsx
```diff
@@ -56,12 +56,7 @@ const BannerCarouselExample = () => {
 
 	return (
 		<div style={{ display: 'flex', flexDirection: 'column' }}>
-			<BannerCarousel id="interactive">
-				{banners.map((banner) => {
-					const BannerComponent = banner.component;
-					return <BannerComponent key={banner.id} id={banner.id} />;
-				})}
-			</BannerCarousel>
+			<BannerCarousel id="interactive" banners={banners} />
```
dani-j-d, line 59: Also, this so much neater...

dani-j-d (approved):
I was able to successfully test! Looks and works great! LGTM!

PR #289, opened 2022-11-23
security: add resolution for loader-utils to resolve dependabot

adamperrry:
PR to update Storybook and add a resolution for `loader-utils` to fix Dependabot alert


dani-j-d (approved):
Built, started, and tested well. LGTM!

PR #293, opened 2022-12-15
fix: move toast styles to top of head

adamperrry:
The `injectStyles()` method provided by `react-toastify` injects the stylesheet at the bottom of `head` via `appendChild`. This is an issue, because in Core, the styles are added via a `link` tag before the injected styles, so the global `Toastify` style overrides are not taking precedence. 

I have opened an issue with the author of that repo [here](https://github.com/fkhadra/style2js/issues/1), but until we get the ability to change that injection method, I have created a workaround to essentially move the offending stylesheet up to the top of `head`. 

The `try/catch` is needed because some style sheets are inaccessible, and trying to access them will throw an error.

### DIFF
--- a/src/components/alerts/ToastContainer/ToastContainer.tsx
+++ b/src/components/alerts/ToastContainer/ToastContainer.tsx
@@ -2,7 +2,28 @@ import React from 'react';
 import { toast as toastDefault, ToastContainer as ToastContainerDefault, ToastContainerProps } from 'react-toastify';
 import { injectStyle } from 'react-toastify/dist/inject-style';
 
-injectStyle();
+const injectToastStyles = () => {
+	injectStyle();
+	let toastStyleSheet: CSSStyleSheet | undefined;
+	// eslint-disable-next-line no-restricted-syntax
+	for (const sheet of document.styleSheets) {
+		try {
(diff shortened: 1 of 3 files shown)


src/components/alerts/ToastContainer/ToastContainer.tsx
```diff
@@ -2,7 +2,28 @@ import React from 'react';
 import { toast as toastDefault, ToastContainer as ToastContainerDefault, ToastContainerProps } from 'react-toastify';
 import { injectStyle } from 'react-toastify/dist/inject-style';
 
-injectStyle();
+const injectToastStyles = () => {
```
austin-redick, line 5: Do you need the try/catch if all the catch does is continue? Maybe just this would be fine:
```javascript
	for (const sheet of document.styleSheets) {
		const rules = sheet?.cssRules;
		const node = sheet?.ownerNode;

		if (rules?.[0].cssText.includes(':root') && rules?.[1].cssText.includes('.Toastify')) {
			node?.remove();
			document.head.insertBefore(node as Node, document.head.firstChild);

			break;
		}
	}
```

adamperrry: I do, because some style sheets aren't accessible, and if you try to access `cssRules` on an inaccessible sheet, it throws an error and blows up :/

austin-redick (dismissed):

nickcernis (approved):

PR #296, opened 2023-01-23
feat: type out IReactComponentProps more accurately and add onContextMenu

adamperrry:
This PR adds better types to `IReactComponentProps`, specifically the handler functions, by pulling in types from React itself. Interfaces extending this interface can now optionally pass a generic type for the kind of HTMLElement, but just `HTMLElement` should be good for the type. I also added in `onContextMenu`. 

This approach of adding what we need when we need it is I think better than just extending `React.HTMLProps`, because that's a lot of props that we sometimes use the names of in our own components but with different types, so it would throw incorrect overload errors.

I also linted the `AccordianItem` component.


austin-redick (approved):
LGTM. Took me a minute to grok all the formatting/linting changes.

PR #297, opened 2023-01-25
feat: extend accordian item to hide the header and expand if header is hidden

adamperrry:
For site group work we need to be able to hide the Accordian header. When the header is hidden, we want the item to be expanded. 

This PR adds an optional prop for doing that!


dani-j-d (approved):
Builds and works great! Tested and header does disappear when noHeader prop is passed, as intended. LGTM! My only hold-up is the lack of an example or any documentation/context in Storybook, but, otherwise, good to go.

PR #298, opened 2023-01-26
feat: Add ThreeDot button component

adamperrry:
For the Site Group work, we need the standalone three dot button from the `ContextMenu` component in order to trigger a system menu. This PR pulls that out into its own component.

I also use it in the `ContextMenu` component and just linted the `PrimaryButton` component.


austin-redick (dismissed):
LGTM

austin-redick (approved):

PR #299, opened 2023-01-26
feat: use IconButton in AccordianItem and spread rest of props

adamperrry:
We have a new IconButton component that handles placing an icon in a button tag, along with hover and active color states.

This commit pulls out the icon button render logic into a separate function and simplifies it a bit.

I also spread the ...restProps onto the wrapping div, giving us things like `onClick`, etc.

### DIFF
--- a/src/components/layout/Accordion/AccordionItem.tsx
+++ b/src/components/layout/Accordion/AccordionItem.tsx
@@ -23,7 +23,7 @@ const animationStateClasses = {
 interface PropsBase extends ILocalContainerProps {
 	children: (React.ReactNode & { props?: any[] }) | null;
 	clickArea?: 'icon' | 'all';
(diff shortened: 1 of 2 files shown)


src/components/layout/Accordion/AccordionItem.tsx
```diff
@@ -22,6 +23,7 @@ const animationStateClasses = {
 interface PropsBase extends ILocalContainerProps {
 	children: (React.ReactNode & { props?: any[] }) | null;
 	clickArea?: 'icon' | 'all';
+	/** Defaults to "true" for default carot icons. Set to "false" for no icon, set to ReactNode for custom icon */
```
nickcernis, line 26: Tiny typo, otherwise looks good!

```suggestion
	/** Defaults to "true" for default caret icons. Set to "false" for no icon, set to ReactNode for custom icon. */
```

nickcernis (approved):

nickcernis (approved):

PR #300, opened 2023-01-30
feat: site group tweaks

adamperrry:
These changes are necessary following feedback from Tyson on the Site Group work.

Includes:
- Accessibly changing the `AccordionItem` header from a `button` to a `div` when the `clickArea` is `"all"`.
- Optionally stopping the propagation of a button's `onKeyDown` handler via a new prop
- Gray mode color tweaks for `IconButton` (the only button using this color mode)


dani-j-d (approved):
Built and tested and it the changes are there without errors or issues! LGTM!

PR #302, opened 2023-02-13
feat: add props for custom classnames on the accordion item content and summary

adamperrry:
Having the ability to pass a classname to these different parts will remove the need for global styles in Core. Also just best practice.


austin-redick (approved):
LGTM

PR #303, opened 2023-02-20
feat: add highlightText option to BasicInput

adamperrry:
Also exports the BasicInput props and adds an onKeyDown event to IReactComponentProps.


austin-redick (approved):
Looks good

PR #304, opened 2023-02-23
feat(LOC-3187): clean up IconCheckbox, Popup, and default outline for focus state

adamperrry:
Cleans up the IconCheckbox and adds a focus state for keyboard navigation.

Changes the default outline for focused elements to $green (we've been wanting that for awhile)

Correctly selects any focused elements. I think Alexa had accidentally put an ampersand which was only selecting the body, not children.

### DIFF
--- a/src/components/inputs/IconCheckbox/IconCheckbox.scss
+++ b/src/components/inputs/IconCheckbox/IconCheckbox.scss
@@ -10,7 +10,7 @@
 	&:not(.IconCheckbox__Disabled) {
 		// mouse down
 		&:active {
-			transform: scale(0.9);
+			transform: scale(0.98);
 		}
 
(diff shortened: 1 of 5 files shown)


bgturner (approved):
Approved -- This does what it says!

I was really trying to get a better understanding of why this component seems overly complicated to me.

I'm curious what Crum's comment of

> // state to track temporarily disabling hover to make the checked/unchecked effect play nicer with hover

is referring to, because I was able to get what I think is the general functionality of checked/hover/disabled styling using the current markup and CSS pseudo classes:

https://i.getf.ly/E0uX2mYP

Anyway, nothing blocking with this PR, I'm mostly curious as to why some of these components seem to rely on JS so heavily for things that could be done with CSS.

PR #308, opened 2023-02-28
feat: Close ContextMenu when clicking a clickable item

adamperrry:
This PR adds a method for closing a tooltip via a custom JS event. This is different than using the hideTooltip prop because we get the animations. 

Using hideTooltip and forceShow to mimic a controlled tooltip loses animations and the ability to still get the trigger, such as a click. Also, using both at the same time creates some wonkiness in the internal state (i.e. closing via the prop then opening and trying to close via the trigger doesn't work right away).

This is a workaround inspired by our `react-toastify` library - a JS event will dispatch state updates to the tooltip stage, triggering the closing animations. We also need to remove listeners for clicks on that event. It works well!

Finally, we close the ContextMenu on clicking any clickable item, something we've been wanting for a while. This more closely mimics real contextMenus. This also opens the possibility for closing the menu on other events, such as scroll.

### DIFF
--- a/src/components/menus/ContextMenu/ContextMenu.tsx
+++ b/src/components/menus/ContextMenu/ContextMenu.tsx
@@ -35,7 +35,7 @@ const ContextMenu = (props: IContextMenuProps) => {
 	const { className, classNameList, classNameListItem, items, onShow, onHide, noBG, bgOnHover, id, ...otherProps } =
 		props;
 
-	const menuId = useRef(`${id ?? 'contextMenu'}-${shortid.generate()}`);
+	const menuId = useRef(id ?? `${'contextMenu'}-${shortid.generate()}`);
 
 	const [isShowing, setIsShowing] = React.useState(false);
 
--- a/src/components/menus/ContextMenu/ContextMenu.tsx
+++ b/src/components/menus/ContextMenu/ContextMenu.tsx
@@ -1,8 +1,10 @@
 import * as React from 'react';
 import classnames from 'classnames';
+import shortid from 'shortid';
+import { useRef } from 'react';
 import styles from './ContextMenu.scss';
 import { FunctionGeneric } from '../../../common/structures/Generics';
-import { Tooltip, TooltipProps } from '../../overlays/Tooltip/Tooltip';
+import { Tooltip, TooltipProps, hideTooltip } from '../../overlays/Tooltip/Tooltip';
 import { TextButton } from '../../buttons/TextButton/TextButton';
 import { ThreeDotButton } from '../../buttons/ThreeDotButton/ThreeDotButton';
 
@@ -16,7 +18,7 @@ export interface IMenuItem {
 	type?: 'normal' | 'separator';
 }
 
-export interface IContextMenuProps extends Omit<TooltipProps, 'content'> {
+export interface IContextMenuProps extends Omit<TooltipProps, 'content' | 'useClickInsteadOfHover'> {
 	/** className for the list items' container */
 	classNameList?: string;
 	/** className for an individual list item */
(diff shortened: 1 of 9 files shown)


src/components/menus/ContextMenu/ContextMenu.tsx
```diff
@@ -30,26 +32,22 @@ export interface IContextMenuProps extends Omit<TooltipProps, 'content'> {
...
-		onHide,
-		noBG,
-		bgOnHover,
-		useClickInsteadOfHover,
-		...otherProps
-	} = props;
+	const { className, classNameList, classNameListItem, items, onShow, onHide, noBG, bgOnHover, id, ...otherProps } =
+		props;
+
+	const menuId = useRef(`${id ?? 'contextMenu'}-${shortid.generate()}`);
```
bgturner, line 38: Do we always want to randomly generate a hash?

If I understand this correctly, the way this is written we will always get some sort of random id:
```typescript
<ContextMenu
	id="something"
	items={[
        // ...
    ]) />
// will give <div class="..." id="something-PrzwA7yje">...</div>
<ContextMenu
	items={[
        // ...
    ]) />
// will give <div class="..." id="contextMenu-PrzwA7yje">...</div>
```


But since `id` is a Typescript `HTMLElement` and therefore just a regular HTML attribute -- it makes sense to me that we would want to be able to explicitly set an `id` and fallback to a randomized string if none is passed. The main reason that comes to my mind would be ease of testing, but there's probably other reasons. I imagine something like:


```typescript
<ContextMenu
	id="something"
	items={[
        // ...
    ]) />
// should give <div class="..." id="something">...</div>
<ContextMenu
	items={[
        // ...
    ]) />
// should give <div class="..." id="contextMenu-PrzwA7yje">...</div>
```

What do you think?

bgturner (changes requested):
A couple of questions in the code, if you could respond.

I'm also curious -- it seems like we are creating a custom JS event, so should we be able to trigger a tooltip remotely -- is that correct?

I was messing around in the `Tooltip.stories.mdx` file to see if I could trigger a tooltip from a sibling button firing an `onClick` event, but I wasn't able to get it working in the 20mins I looked. Am I missing something simple, or am I trying to do something that this component isn't capable of?

If we should be able to remotely control a tooltip, I'd love to see an example, either in `Tooltip.stories.mdx` or `Tooltip.test.tsx`.

src/components/overlays/Tooltip/Tooltip.tsx
```diff
@@ -164,13 +164,21 @@ const reducerInitialState: ReducerState = {
...
  * used to eliminate events firing for each transition property being used
  */
-const useTooltipStage = (
-	isTransitionEnd: boolean,
-	isHoverTarget: boolean,
-	isHoverPopper: boolean,
-	hideDelay: number | undefined,
-	showDelay: number | undefined
-) => {
+const useTooltipStage = ({
```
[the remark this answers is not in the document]
adamperrry: Stage here just refers to the stage of opening or closing the tooltip is currently in. It's used to handle the complex positioning/animations. I think stage is a fine term for it. 

Again, this is pretty complex component, and it'd be nice if we didn't have to do it this way, but I can't rewrite such a ubiquitous, working component

src/components/menus/ContextMenu/ContextMenu.tsx
adamperrry: Yeah whenever I do this I always forget parent components might want to be able to explicitly know the ID of element, specifically for playwright tests, yeah. 

I think I'll do what you said and fallback to a unique hash but default to the passed ID. My main concern is there being multiple context menus with the same ID, but that's more on devs to realize not to do that.

adamperrry: Also, we'd pretty much only every have one contextmenu open at a time, because clicking outside of one would close the other

bgturner (approved):
Looks good! Thanks for clarifying things and providing that example to Storybook!

PR #311, opened 2023-03-07
fix: increase drag region

adamperrry:
Does what it says!

I thought we'd need to disallow dragging on any buttons that were underneath the drag area, but something might have changed during an electron upgrade, because now we can click the buttons under the drag area while still dragging there too. I remember that not being the case when I originally changed this to 10px.

### DIFF
--- a/src/styles/_partials/_mixins.scss
+++ b/src/styles/_partials/_mixins.scss
@@ -84,7 +84,7 @@
 	top: 0;
 	left: 0;
 	right: 0;
-	height: 10px;
+	height: 30px;
 	width: 100%;
 }
 


nickcernis (approved):
Looking good.

Only comment is that 20px still feels a bit short to me. (VS Code seems to have a draggable target of ~32px, Slack/Chrome are ~45px.) But it's better than it was!

Tyson mentioned trying 30px if we can in the Jira story. Did that not work out?

| Before | After |
| - | - |
| Screenshot 2023-03-08 at 10 19 08 | Screenshot 2023-03-08 at 10 18 05 |

PR #313, opened 2023-04-05
fix: close open dependabots

adamperrry:
Includes easy fixes for `dns-packet` and `webpack,` and resolution fixes for `terser` and `json5`.

Was hoping an upgrade to storybook would fix terser, but it didn't. Also, we're running out of upgrade space for storybook, looks like version 7.0 is out, but that'll require upgraded node. Should be a good time to re-evaluate our Storybook instance if we bring this inside Local Core.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,9 +126,9 @@
     "ts-node": "^8.2.0",
     "typescript": "^4.7.4",
     "url-loader": "^4.1.1",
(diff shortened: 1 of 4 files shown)


bgturner (approved):
Looks good! Tests still pass and storybook still opens!

PR #315, opened 2023-05-17
feat: remove global Downloader css and refactor DownloaderItem and DownloaderOverlay

adamperrry:
Introduces a `DownloaderOverlay` wrapper because we were implementing `DownloaderItem`s with the same wrapper in two places in core. Also, we were using global css, which is always frustrating. Just moves the css into a scoped file to make this easier to work on.

Also, I tweaked the way things are styled and set up, so that the progress bar doesn't jump around when the content above or below it changes. I probably could've achieved this with absolute positioning, but I chose to just make the top and bottom sections equal heights and centering everything in a flex container, like a div sandwich.

Additionally, changes to a functional component and cleans everything up, adds a story for `DownloaderOverlay`.

Removes the global `SiteInfo.scss` file because nothing in core was actually using these globally-scoped class names.


src/components/loaders/DownloaderItem/Downloader.scss
```diff
@@ -0,0 +1,58 @@
+@import '../../../styles/_partials/index';
+
+.DownloaderOverlay {
+	z-index: 10;
```
adamperrry, line 4: This change to z-index fixes a bug where some modals were appearing behind the overlay. E.g. when adding a site then trying to delete it.

austin-redick (approved):
LGTM! Thanks for re-working the old class components.

PR #320, opened 2023-09-13
feat: add maxHeightOffest option for Combobox

adamperrry:
I'm going to try and add this component to Cloud backups for backup selection, but the dropdown gets hidden by the stepper, and the top component scrolls. This solution adds an option to adjust the max height, which is dynamically set to 75px above the bottom of the window and is reset on window resize.

### DIFF
--- a/src/components/inputs/Combobox/Combobox.tsx
+++ b/src/components/inputs/Combobox/Combobox.tsx
@@ -125,6 +125,8 @@ export interface IComboboxProps extends ILocalContainerProps {
 	invalid?: boolean;
 	/** Message shown below text input when combobox is invalid */
 	invalidMessage?: string;
+	/** Offset for the max height of the combobox options dropdown. The max height defaults to 75px above the bottom of the window. */
+	maxHeightOffset?: number;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/components/inputs/Combobox/Combobox.tsx
```diff
@@ -144,6 +146,7 @@ const Combobox = (props: IComboboxProps) => {
 		onChange,
 		options = {},
 		optionHeight = 's',
+		maxHeightOffest = 0,
```
bgturner, line 149: Should this be `...Offset` instead of `...Offest`?

bgturner (changes requested):
Looks like a typo.

I'm also having trouble visualizing the need for this. Can you add a section to Storybook that highlights the issue that this is trying to solve so I can better understand the need for this option?

src/components/inputs/Combobox/Combobox.tsx
adamperrry: Yep! And as to why this is necessary - we dynamically set the maxheight for the dropdown based on the Window height (so that it doesn't overflow the window). It stops like 75 px above the window bottom. But in the case of Cloud Backups, we don't really want to cover up the stepper, because it looks bad. So this allows us to decrease the max height.

nickcernis (approved):

PR #326, opened 2023-10-03
fix: add .nvmrc at 18.15.0

bgturner:
I was getting an [error in Azure](https://dev.azure.com/localbyflywheel/Local%20by%20Flywheel/_build/results?buildId=6460&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=bd05475d-acb5-5619-3ccb-c46842dbc997&l=20) because the environment was 14.17.0 but nanoid requires a newer version.

This commit adds an `.nvmrc` and updates azure to use that value so that we have a more predictable dev environment that matches the CI/CD env.

```
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
error nanoid@5.0.1: The engine "node" is incompatible with this module. Expected version "^18 || >=20". Got "14.17.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
##[error]Bash exited with code '1'.
```


azure-pipelines.yml
```diff
@@ -14,7 +14,8 @@ pool:
 steps:
 - task: NodeTool@0
   inputs:
-    versionSpec: '14.17.0'
+    versionSource: 'fromFile'
+    versionFilePath: '.nvmrc'
```
mike-day, line 18: Nice update! It makes sense to peg this to `.nvmrc`

mike-day (approved):

PR #329, opened 2023-10-18
Upgrade Storybook from v6 to v7.5

bgturner:
These changes were mostly a result of running the automated script and following the prompts:

```shell
npx storybook@latest upgrade
```

There was an issue with Storybook not seeming to recognize JSX within Typescript. It sort of sounded like this issue: https://github.com/storybookjs/storybook/issues/22357 

Because our `.babelrc` file wasn't too complex, I opted to delete it and let Storybook re-generate it with:

```shell
npx storybook@latest babelrc
```

Resolves: LOC-5350


nickcernis (dismissed):
Looking good locally.

Screenshot 2023-10-20 at 16 14 40

mike-day:
Not a blocker by any means, but we should consider a change in the way we structure stories to take advantage of the expanded autodocs features in storybook.

For example, refactoring the `Alert` story  results in seeing all available props in the docs:

Screenshot 2023-10-20 at 10 04 55 AM

---

For the above example, I actually deleted the MDX file and refactored the `Alert` story as below:

Screenshot 2023-10-20 at 10 07 41 AM

austin-redick (approved):
Overall, LGTM! Personally, I think the prettier de-structuring rule is a bit strict in cases like this; it adds so many lines!

bgturner:
> Overall, LGTM! Personally, I think the prettier de-structuring rule is a bit strict in cases like this; it adds so many lines!

I'm still formulating an opinion on it. Part of me likes how many lines it makes, because it makes it a lot easier to see all the things you're pulling out and as a result, you get a sort of code-smell for an interface being large.

PR #330, opened 2023-10-19
[LOC-5305] Adjust components for waffle menu / product drawer

nickcernis:
These adjustments were made for the incoming waffle menu PR at https://github.com/getflywheel/flywheel-local/pull/1813. 

In rough order of relevance/impact, this PR:

- Introduces a 'button' type for the `VerticalNavItem` (with new aria props) for use as a nav item that directly triggers the passed 'onClick' function instead of navigating to another page. (The product drawer does not navigate the user to a new page; it presents an app-wide overlay. It may be possible to achieve this with NavLink hacks, but simply calling a function on button press seems much cleaner to me.)
- Gives `VerticalNav` a new z-index of 9 so that it appears above the new product drawer, but below z-index 10 items such as the new site page.
- Allows the Close button `color` to be specified so it is not always green (the product drawer button is styled in a neutral color).
- Exports `ButtonPropColor` at the package level so we can use the type for the Close button color.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@getflywheel/local-components",
-  "version": "17.7.2",
+  "version": "17.7.3-beta.7",
```
kienstra, line 3: Good idea to mark it beta

src/components/modules/VerticalNav/VerticalNav.scss
```diff
@@ -118,4 +119,26 @@
...
+		border: none;
+		cursor: pointer; /* Override 'cursor: default' in base.scss. */
+		width: 100%;
+
+		svg {
+			cursor: pointer; /* Override 'cursor: default' in base.scss. */
+			width: 22px;
+
+			path {
+				cursor: pointer; /* Override 'cursor: default' in base.scss. */
```
kienstra, line 134: Nice comments

src/components/modules/VerticalNav/VerticalNav.tsx
```diff
@@ -86,6 +123,8 @@ export const VerticalNavItem = (props: IProps) => {
 	switch (type) {
 		case 'addsite':
 			return renderTypeAddSite();
+		case 'button':
+			return renderTypeButton();
```
kienstra, line 127: Good idea to add this, it's kind of a Strategy Pattern of render methods.

kienstra (approved):
Hi @nickcernis,
Nice work!

src/components/modules/VerticalNav/VerticalNav.scss
```diff
@@ -118,4 +119,26 @@
...
+			}
+		}
```
adamperrry, line 136: I think you could replace all the `cursor: pointer` with something like:

```scss
	.VerticalNavButton, .VerticalNavButton * {
		cursor: pointer; /* Override 'cursor: default' in base.scss. */
	}

	.VerticalNavButton {
		background: none;
		border: none;
		width: 100%;

		svg {
			width: 22px;
		}

		&:active svg path,
		&:hover svg path,
		&[aria-expanded="true"] svg path {
			@include theme-fill-greendark50-else-white-hover;
		}
	}
```

src/components/modules/VerticalNav/VerticalNav.tsx
```diff
@@ -22,18 +22,35 @@ export const VerticalNav = (props: INavProps) => {
 
 interface IProps extends IReactComponentProps {
 	className?: string;
+	ariaLabel?: string;
+	ariaExpanded?: boolean;
+	ariaControls?: string;
```
adamperrry, line 27: Would it make sense to add these to `IReactComponentProps` so that they're available for all of our components to accept and use? Something like:

```typescript
export default interface IReactComponentProps<T = HTMLElement> {
    // Rest of interface...
	'aria-checked'?: React.HTMLProps<T>['aria-checked'];
	'aria-label'?: React.HTMLProps<T>['aria-label'];
	'aria-expanded'?: React.HTMLProps<T>['aria-expanded'];
	'aria-controls'?: React.HTMLProps<T>['aria-controls'];
```

It's could be nice to be able to capture those props in a `...restProps` while deconstructing and then spread them all at once on the applicable element. Not a blocker though, just an idea!

src/components/modules/VerticalNav/VerticalNav.tsx
```diff
@@ -71,6 +88,26 @@ export const VerticalNavItem = (props: IProps) => {
...
+			<button
+				className={styles.VerticalNavButton}
+				type="button"
+				aria-label={ariaLabel || tooltip || undefined}
+				aria-expanded={ariaExpanded}
+				aria-controls={ariaControls || undefined}
+				onClick={() => {
+					if (action) {
+						action();
+					}
```
adamperrry, line 102: I think you could do away with the `action` prop and just grab `onClick` off of `props` (it exists in `IReactComponentProps`). And I think you wouldn't need the check, just:

```typescript
onClick={onClick}
```

adamperrry (changes requested):
Looking good! Just a few thoughts. Also, could we add an example of the `button` type to the `VerticalNav.stories.mdx`?

src/components/modules/VerticalNav/VerticalNav.tsx
nickcernis: Done!

src/components/modules/VerticalNav/VerticalNav.tsx
nickcernis: Ah, nice idea, thanks for suggesting. Done!

nickcernis:
Thanks for the reviews, @kienstra and @adamperrry.

@adamperrry I made your suggested changes and added an example of the 'button' type to Storybook. This is ready for another look!

adamperrry (approved):
Very nice! Good to have examples of aria usage popping up

adamperrry (approved):

PR #331, opened 2023-10-23
Jb/fix/misc a11y

dreamwhisper:
## Summary
 - Ensures icons are not read out as `image` on buttons and in IconCheckbox. (may depend on operating system, assistive technology used)
 - Adds a `label` prop to IconCheckbox. At present, a screen reader reads it only as a checkbox with no indication what it is for.

## Technical
Should be tested with [this branch in flywheel-local](https://github.com/getflywheel/flywheel-local/tree/jb/fix/simple-a11y).

## Reference

- https://wpengine.atlassian.net/jira/software/c/projects/GF/boards/441?selectedIssue=GF-3875

## Notes
These should be switched out for buttons with proper aria.


kienstra (approved):
Hi @dreamwhisper,
This looks good!

